PR #133, opened 2014-12-29
Nk

mrshirts:
Here's the proposed generalization of the expectations.  Not quite as clean as I would have hoped, and took a while to get it to this state, but does eliminate a lot of the non-vectorized loops.

It appears to be slightly slower for larger N (500K to 2M states), and I have not yet tested the performance on large numbers of states (sampled or unsampled) between the codes.  It may require a bit of tweaking.

I didn't try to run the PMF code through it as well; that would have meant a number of other conditionals that seemed to complicated the code too much.

It may make sense to split out the computePerturbedFreeEnergies functions; that will simplify the inner loop a bit more.


mrshirts:
I've noticed there is a test_mbar.py and test_covariance.py pymbar/tests already, but they don't seem to have any calls.  Should this be set up with a main like in Yank/tests/test_repex.py? Is there another routine that is supposed to call the tests that integrates with CI?  Should tests just raise exceptions and return nothing, like the tests in pymbar/tests/test_mbar_solvers.py? Does _test have a special meaning?  If they are nosetests, why is SkipTest the only nose functionality used?

I'm happy to do all of the heavy lifting for getting the tests running, but I'm still very uncertain as to how you are imagining the testing framework is constructed. The two examples are set up differently, leaving me unclear what sort of framework is envisioned. If you can take the test_mbar.py with a sample function that makes it clear what exceptions to raise and what should be returned, with an explanation of where it's called, then I can go in and write the duplicate functions that test all the functionality. Or, point to documentation that explains exactly how you want the tests structured, such that I can actually run the tests after writing them and know that they are generating the correct output.

kyleabeauchamp:
1.  A `main()` is not required.  Nosetests uses a regex to search for functions and classes with test in the name, then runs them. 
2.  `_test()` is a "helper function" that is used by the several "actual" tests--I've just abstracted it out to share code.
3.  Yes, tests just raise exceptions when an error is found.  If the exception is a `SkipTest`, however, nose knows to pass over that exception.

So in summary, we just need a python module containing a list of functions with "test" in the name.  These functions raise exceptions if something is wrong.  Beyond that, I don't have any requirements.

I know this is a bit different.  We have some (nothing we haven't already discussed here, though) guidelines here (http://mdtraj.org/latest/style.html).  Sklearn seems to use a similar scheme, e.g. (https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/tests/test_dummy.py).

kyleabeauchamp:
In case you are wondering, the *advantage* of this framework is that every time someone adds a feature via pull request, adding a test can be as simple as adding a single function to one of the testing modules.

kyleabeauchamp:
One other thing: in a future pull request, I'm going to make it possible so that users can run the entire testing framework *after* installation via `nosetests pymbar`.  This means that regardless of how people install (conda, pip, setup.py), they will have the ability to run the test suite.

jchodera:
Sorry for being mostly absent from this thread. Still hiking. But I totally
agree that we eventually need a robust and general statistical testing
framework and can work this out soon.

For now, just encoding some basic tests is the way to go, even if they are
rudimentary.

mrshirts:
Thanks for clarifications, that's very helpful.  I'll get something
together by today.

kyleabeauchamp:
Great--thanks for playing along with this.  My goal here is not to make you do a huge amount of heavy lifting, but to make sure you're up to speed on how things are working.

mrshirts:
> A main() is not required. Nosetests uses a regex to search for functions and classes with test in the name, then runs them.

So, there can be no arguments to the test?  For example would a sample have to be regenerated for every test?  Is the way to work around this to generate the data globally?  This could get complicated if one wants to be able to do repeatable tests.

> _test() is a "helper function" that is used by the several "actual" tests--I've just abstracted it out to share code.

So _test() doesn't get called itself even though it has 'test' in the name?

kyleabeauchamp:
Yes, there are ways to share the setup and tear-down code between tests.  So the choices are use that more complex functionality in nose or define globals.  

`_test()` does not get called as a test.

mrshirts:
Something I've brought up a few times here but haven't gotten an answer
on-- one of the most important things I think we may need with
modifications is regression testing, since there can be subtle differences
that appear, that because of the statistical nature of the calculation,
need to be repeated determininstically on the same imput?  What is the
framework you want to be using for regression testing?  I can't find too
much on nosetests on regression testing, since it's designed as a unit test
environment. I can wrap around the harmonic-oscillators.py example easily
enough and compare to existing pickles or test files, but are there other
approaches you are thinking of?

kyleabeauchamp:
We don't have a good answer for that yet--sorry for not answering that more directly.  If you have tests that really cannot be treated as unit tests, I'm happy converting those tests later.  

I suspect the sklearn folks work around this by fixing the RNG seed to achive determinism--it's hacky, I know.

I guess my goal here is "use nosetests style as much as possible."  I don't mean to force a square peg into a round hole.

kyleabeauchamp:
Can you run `nosetests` on your machine and fix the existing tests to work with your changes?  I see the following errors:

```
Doctest: pymbar.mbar.MBAR.computeOverlap ... FAIL
Doctest: pymbar.mbar.MBAR.computePMF ... ok
Doctest: pymbar.mbar.MBAR.computePerturbedFreeEnergies ... ok
Doctest: pymbar.mbar.MBAR.getFreeEnergyDifferences ... ok
Doctest: pymbar.old_mbar.MBAR.__init__ ... ok
Doctest: pymbar.old_mbar.MBAR.computeEntropyAndEnthalpy ... ok
Doctest: pymbar.old_mbar.MBAR.computeExpectations ... ok
Doctest: pymbar.old_mbar.MBAR.computeGeneralExpectations ... ok
Doctest: pymbar.old_mbar.MBAR.computeMultipleExpectations ... ok
Doctest: pymbar.old_mbar.MBAR.computeOverlap ... ok
Doctest: pymbar.old_mbar.MBAR.computePMF ... ok
Doctest: pymbar.old_mbar.MBAR.computePerturbedFreeEnergies ... ok
Doctest: pymbar.old_mbar.MBAR.getFreeEnergyDifferences ... ok
Doctest: pymbar.timeseries.detectEquilibration ... ok
Doctest: pymbar.timeseries.normalizedFluctuationCorrelationFunction ... ok
Doctest: pymbar.timeseries.normalizedFluctuationCorrelationFunctionMultiple ... ok
Doctest: pymbar.timeseries.statisticalInefficiency ... ok
Doctest: pymbar.timeseries.statisticalInefficiencyMultiple ... ok
Doctest: pymbar.timeseries.subsampleCorrelatedData ... ok
Doctest: pymbar.utils._logsum ... ok
tests.test_covariance.test_100x100_oscillators ... ok
tests.test_covariance.test_200x50_oscillators ... ok
tests.test_covariance.test_200x50_exponentials ... ok
tests.test_covariance.test_gas ... SKIP: skip
tests.test_covariance.test_8proteins ... SKIP: skip
Generate test objects and calculate analytical results. ... ok
Draw samples via test object. ... ok
Can MBAR calculate correct free energy differences? ... ok
Can MBAR calculate E(x_n)?? ... FAIL
Can MBAR calculate E(x_n^2)?? ... FAIL
Can MBAR calculate general expectations. ... ERROR
```

kyleabeauchamp:
Looks like the output of some functions have switched from `np.ndarray` to `np.matrix`:

```
======================================================================
FAIL: Can MBAR calculate E(x_n)??
----------------------------------------------------------------------
  File "/home/kyleb/opt/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/kyleb/src/mrshirts/pymbar/tests/test_mbar.py", line 82, in test_mbar_expectation_xkn
    eq(z / z_scale_factor, np.zeros(len(z)), decimal=0)
  File "/home/kyleb/src/mrshirts/pymbar/pymbar/utils_for_testing.py", line 116, in eq
    assert (type(o1) is type(o2)), 'o1 and o2 not the same type: %s %s' % (type(o1), type(o2))
AssertionError: o1 and o2 not the same type: <class 'numpy.matrixlib.defmatrix.matrix'> <type 'numpy.ndarray'>
-------------------- >> begin captured stdout << ---------------------
Harmonic Oscillators

--------------------- >> end captured stdout << ----------------------

======================================================================
FAIL: Can MBAR calculate E(x_n^2)??
----------------------------------------------------------------------
  File "/home/kyleb/opt/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/kyleb/src/mrshirts/pymbar/tests/test_mbar.py", line 99, in test_harmonic_oscillators_mbar_xkn_squared
    eq(z / z_scale_factor, np.zeros(len(z)), decimal=0)
  File "/home/kyleb/src/mrshirts/pymbar/pymbar/utils_for_testing.py", line 116, in eq
    assert (type(o1) is type(o2)), 'o1 and o2 not the same type: %s %s' % (type(o1), type(o2))
AssertionError: o1 and o2 not the same type: <class 'numpy.matrixlib.defmatrix.matrix'> <type 'numpy.ndarray'>
```

IMHO, I think it's best to use np.ndarray as input and output whenever possible--it's really the better supported class in most cases.  I'd vote to use `np.matrix` internally when we need lots of readable matrix operations.

pymbar/mbar.py
```diff
@@ -291,6 +291,55 @@ def getWeights(self):
...
+                      = \int (p_i p_j /\sum_k N_k p_k) dq^N
+
+        Multiplying elementwise by N_i, the elements of row i give the probability
+        for a sample from state i being observed in state j.
+
+        Examples
+        --------
+
+        >>> from pymbar import testsystems
+        >>> [x_kn, u_kn, N_k] = testsystems.HarmonicOscillatorsTestCase().sample(mode='u_kn')
```
kyleabeauchamp: This line should have an extra variable `s_n` that represents the state of origin.  Our testsystems now return that as well.  E.g.:

```
(x_n, u_kn, N_k, s_n) = testsystems.HarmonicOscillatorsTestCase().sample(mode='u_kn')
```

kyleabeauchamp:
So the changes here now have `computeExpectations` returning a list containing a np.ndarray and an np.matrix:

```
In [6]: mbar.computeExpectations(x_n)
Out[6]: 
[array([ 0.11798645,  0.79207608,  1.95702241,  2.93389592,  4.01794273]),
 matrix([[ 0.14793763,  0.12534317,  0.07345266,  0.0561383 ,  0.03192595]])]
```

IMHO we should just stick to np.ndarray for the output.

mrshirts:
The latest post just fixed the matrix/array issues.

mrshirts:
Updated testing, validation and examples.
    
    * updated README.mds (some more revision likely still needed)
    * Added QQ plots and Anderson-Darling statistical test
    for more comprehensive statistical tests.
    * added examples/ folder with harmonic-oscillators.py and
      harmonic-oscillators-distributions.py and sample output
    * in tests/ fixed test_mbar_solvers.py, and added a large number of new test_mbar.py tests, including at least one test for 
    * cleaned up some output: now only ndarrays are output. not np.matrices.
    * a fix in the PMF zero state referencing
    * Some reconfiguring of  the pymbar/testsystems.py for test_mbar.py

mrshirts:
> This line should have an extra variable s_n that represents the state of origin

Fixed in separate commit.

kyleabeauchamp:
Thanks, I'm now looking over the updated changes.

mrshirts:
FYI: I'm getting the following errors in most cases when I initialize MBAR
/Users/mrshirts/anaconda/lib/python2.7/site-packages/scipy/optimize/_minimize.py:314: RuntimeWarning: Method L-BFGS-B does not use Hessian information (hess).
  RuntimeWarning)
/Users/mrshirts/anaconda/lib/python2.7/site-packages/scipy/optimize/minpack.py:236: RuntimeWarning: xtol=0.000000 is too small, no further improvement in the approximate
  solution is possible.
  warnings.warn(msg, RuntimeWarning)
/Users/mrshirts/work/papers/MBARFORDUMMIES/git/pymbar/pymbar/mbar.py:265: RuntimeWarning: divide by zero encountered in log
  self.Log_W_nk = np.log(mbar_solvers.mbar_W_nk(self.u_kn, self.N_k, self.f_k))

kyleabeauchamp:
I've made those a separate issue.  The first two are just a verbosity issue.  I think the last one is fine, but I'll check.

mrshirts:
OK, I'll be out for a couple of hours.  Hopefully, the rest will just be minor cleanup.

It's possible there will be some slowdown with the generalized expectations code, which might need to be reverted. I haven't timed very large number states for computeExpectations or computePerturbedFreeEnergies -- it won't effect the timing to initialize MBAR itself.

The Anderson-Darling functionality will allow us to write some more stringent statistical tests soon, which I may be able to do as we get the expectations paper written, though not this week.

kyleabeauchamp:
Sounds good.  I'll probably merge this once I read over things one more time.  Thanks!

jchodera:
Thanks again, everyone!

Will start a discussion about more extensive testing framework in another thread.

PR #143, opened 2015-01-05
Adding effective sample functionality

mrshirts:
uses the weight matrix to calculate an effective sample number collected at each state (i.e. MBAR gives more effective samples at each state).  This is particularly useful to see when the effective sample number is very low  (< 10), meaning that very few samples are contributing to the ensemble averages.

### DIFF
--- a/examples/harmonic-oscillators.py
+++ b/examples/harmonic-oscillators.py
@@ -531,6 +531,41 @@ print "Overlap scalar output"
 print O
 
 print "============================================"
+print "    Testing computeEffectiveSampleNumber    "
+print "============================================"
+
+N_eff = mbar.computeEffectiveSampleNumber(verbose = True)
+print "Effective Sample number"
+print N_eff
+print "Compare stanadrd estimate of <x> with the MBAR estimate of <x>"
+print "We should have that with MBAR, err_MBAR = sqrt(N_k/N_eff)*err_standard,"
+print "so standard (scaled) results should be very close to MBAR results."
+print "No standard estimate exists for states that are not sampled."
+A_kn = x_kn
+(val_mbar, err_mbar) = mbar.computeExpectations(A_kn)
+err_standard = numpy.zeros([K],dtype = numpy.float64)
+err_scaled = numpy.zeros([K],dtype = numpy.float64)
+
+for k in range(K):
+  if N_k[k] != 0:
+    # use position
+    err_standard[k] = numpy.std(A_kn[k,0:N_k[k]])/numpy.sqrt(N_k[k]-1)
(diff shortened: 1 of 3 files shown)


kyleabeauchamp:
Overall looks good to me.  I have a few ideas on how to streamline the code, but I'll just do that myself as a separate PR.

jchodera:
I'm not sure I like the term "effective sample number".  Maybe "effective state contribution" or "effective state contribution weight"?  The effective number of samples, in terminology that is often used in computing statistical errors, is the effective sample size `N_eff`, where `stderr = std(A)/sqrt(N_eff)`.

mrshirts:
> The effective number of samples, in terminology that is often used in computing statistical errors, is the effective sample size N_eff, where stderr = std(A)/sqrt(N_eff).

That's exactly what this is.  Ensemble average from MBAR behave (roughly, since the Kish measure isn't perfect) as if the variance is divided by sqrt(N_eff), not std(N) from that state.

jchodera:
My problem is with the function you added called `computeEffectiveSampleNumber()`.  

I want to make sure the function name, the calculation, and the docstring are all conceptually consistent.

First, the docstring contains inconsistencies:
* The [function description](https://github.com/choderalab/pymbar/blob/master/pymbar/mbar.py#L296-L298) states:
```
        Compute the effective sample number of each state;
        essentially, an estimate of how many samples are contributing to the average
        at given state.  See pymbar/examples for a demonstration.
```
while the [documentation for the return value](https://github.com/choderalab/pymbar/blob/master/pymbar/mbar.py#L307-L308) says something different:
```
                estimated number of states contributing to estimates at each 
                state i.
```
And then there's [this thing I don't understand](https://github.com/choderalab/pymbar/blob/master/pymbar/mbar.py#L300-L302) which doesn't seem to actually be returned:
```
        It also counts the efficiency of the sampling, which is simply the ratio
        of the effective number of samples at a given state to the total number
        of samples collected.
```
The actual thing returned simply sums up the total inverse-squared-weight contributing to an estimate for each target thermodynamic state.  It's not clear to me that this is the same as the `(std(A)/stderr)^2`.  Instead, this seems like something more like an indicator of whether very few individual samples are contributing to the average, rather than an estimate of the *effective sample size*.  Is there some connection here that I'm missing?

mrshirts:
>        Compute the effective sample number of each state;
>        essentially, an estimate of how many samples are contributing to the average
        at given state.  See pymbar/examples for a demonstration.
while the documentation for the return value says something different:

>                estimated number of states contributing to estimates at each 
>               state i.

Right, this second one should say "estimated number of samples contributing to estimates at each state i".

>It also counts the efficiency of the sampling, which is simply the ratio
>        of the effective number of samples at a given state to the total number
>        of samples collected.

It does the count, but prints it in the verbose.  It doesn't say that it returns it.

Issue #139, opened 2015-01-03
Proposed deletion computePMF_states


mrshirts:
It has less functionality than computePMF, and I don't see any advantages to it.

jchodera:
Agreed. This was experimental, and existed only because it had potential advantages in numeric stability or speed. Feel free to do this.

kyleabeauchamp:
Done

PR #144, opened 2015-01-05
Remove PMF_states code

mrshirts:
Like it says.  examples and code removed.  See https://github.com/choderalab/pymbar/issues/139


jchodera:
Looks good to me.

PR #148, opened 2015-01-05
Eliminated variable length returns

### LINKED ISSUE #22: Variable length function outputs
This is a minor style comment, but a lot of our functions look like this:

```
[...]

return returns
```

Where returns is a list whose length various depending on a bunch of optional arguments to the function.

To me, this style is unclear.  Whenever possible, I would prefer for our functions to return a fixed number of variables--and returned by name:

```
return u_kn, f_i, kitchen_sink

### PR DESCRIPTION
**Pull request by kyleabeauchamp**
Fixes #22 

I left `computeExpectationsInner()` as is (e.g. dictionary), as that function is primarily geared intended for internal use.

### CODE REVIEW

**Comment by jchodera** [comment]
If we aren't going to use the third argument, I think we can just omit it and nothing will complain:
https://github.com/choderalab/pymbar/pull/148/files#diff-6d49c4d7bec5063b52a900e8455a638bR135

For example, could we change
```
(Deltaf_ij_estimated, dDeltaf_ij_estimated, _theta) = mbar.getFreeEnergyDifferences()
```
to
```
(Deltaf_ij_estimated, dDeltaf_ij_estimated) = mbar.getFreeEnergyDifferences()
```
?

**Comment by jchodera** [comment]
Nope, my bad.  That doesn't work. 
```
>>> def foo():
...    return (1, 2, None)
... 
>>> [x,y] = foo()
  File "<stdin>", line 1, in <module>
ValueError: too many values to unpack
```
Odd---I thought it did.

**Comment by jchodera** [comment]
I see now.  My earlier example tricked me:
```
>>> def foo():
...    return 1, None
... 
>>> [a,b] = foo()
>>> a = foo()
```
Here, the last line assigns `a` to the tuple `(1, None)`.

mrshirts:
It's less important to return theta at this point, because now if one wants
to return the covariance matrix, one can go down just one level, to the
inner loop code, instead of writing the routine from scratch.

Also, the covariance of differences code makes it easier to calculate the
uncertainties of sums and differences without theta. Note that that
function still needs a good example -- I'll ported that from another code
branch, but it doesn't have examples yet.

kyleabeauchamp:
OK, if desired we can do a separate pull request to remove the `theta` returns.

PR #151, opened 2015-01-08
Updated README, removed cython dependency

kyleabeauchamp:


jchodera:
cython is dead---log live cython!

PR #152, opened 2015-01-09
Switched to int64, moved logsumexp to utils.

kyleabeauchamp:
1.  Switched int32 to int64.  I'm worried about someone sending a billion samples into mbar, which is bound to happen at some point.  
2.  For consistency with the old code, I moved `logsumexp()` from `mbar_solvers.py` to `utils.py`.  
3.  I've switched calls to the old `_logsum()` to use `logsumexp()`
4.  For now, I've left `_logsum()` where it is and use it as a reference in our unit tests comparing to `logsumexp()`

### DIFF
--- a/pymbar/bar.py
+++ b/pymbar/bar.py
@@ -49,8 +49,7 @@ __license__ = "LGPL 2.1"
 #=============================================================================================
 import numpy as np
 import numpy.linalg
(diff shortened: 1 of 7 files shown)


mrshirts:
Looks good to me.

PR #156, opened 2015-01-11
Improved harmonic_oscillators.py example

mrshirts:
Runs both 1d and 2d PMF calculations, using reusable code.  For some other PMF tools, it's useful to have a 1D example in the example set.


mrshirts:
Also, it moves to NxK instead of NxKxK data representation.

PR #157, opened 2015-01-11
Move examples to pymbar repository

kyleabeauchamp:


jchodera:
This is a good opportunity to consider whether we need ALL of the files in
this before we commit it.

Are there any inessential data output files we can omit?

kyleabeauchamp:
So if there are non-essentially files, I think we should delete them via separate pull request to `pymbar-examples`.  Then I will copy the files into `pymbar` and file a new PR.

mrshirts:
Deleting files before a move is better, because if they are deleted after,
they stay in the repository.

mrshirts:
alchemical-free-energy/gromacs/output_11steps_skip_lambda should be deleted.  It's no longer used.

The alchemical-free-energyamber/data/ti00*/ti00*.out files together take up 30M.  I wonder if they should be zipped or something in the repository.  It's only about 200 ps of data (and only output every 10 ps) so we probably can't truncate it.

constant-force-optical-trap/original_data and /processed_data contain the second most data, but I don't know what can be eliminated there.

kyleabeauchamp:
I think bzipping some of the outputs is a reasonable idea, given that we only need to check the reference output very infrequently.

mrshirts:
It would probably best to do that before merging, so the full files + the bzipped files aren't in the repository.  Also, does Windows have an issue with bzipped files?

kyleabeauchamp:
We can do zip files if we want, or tar.gz.  AFAIK, winrar supports bz2 (http://en.wikipedia.org/wiki/WinRAR)

I think we have two steps:

1.  File pull requests to pymbar-examples deleting unused / obsolete files
2.  File a pull request to pymbar-examples putting some (or all) of the output files in some form of tarballs.
3.  I file a pull request to pymbar copying in the files.

kyleabeauchamp:
I'm going to close this pull request for now as we apparently haven't decided which files we want to delete versus tarball versus include in plaintext.

mrshirts:
I removed and compressed files to result in 32 M more space.

PR #159, opened 2015-01-14
Stefano/time series fix

smcantab:
removed bug in loop of detectEquilibration_binary_search, added kwarg "bs_nodes" to vary number of initial nodes, now the function adapts the number of nodes to refine result until this is less than minimum stable number of nodes, which is 4 (this way the final error will be smaller even if a large number of nodes are used at the start, calculation stops when )

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -10,7 +10,7 @@ requirements:
     - scipy
     - setuptools
     - numexpr   
-    - six 
+    - six
   run:
     - python
     - numpy
@@ -21,7 +21,8 @@ requirements:
 test:
   requires:
     - nose
-    - setuptools    
+    - setuptools
+    - statsmodels    
   imports:
     - pymbar
   commands:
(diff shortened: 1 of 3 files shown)


smcantab:
actually, we want to exclude the last point in the grid only when it is the size of the time series (thus triggering the bug), otherwise we can include it. Let me try something else, I'll commit a change.

jchodera:
This is great!

I think we need one more minor addition: We should add the `statsmodel` package to the `requirements:` part of the `test:` section of the development conda recipe:
https://github.com/choderalab/pymbar/blob/master/devtools/conda-recipe/meta.yaml#L22-L24

That way, travis will also install the `statsmodel` package and run your test automatically!

jchodera:
Odd.  The `statsmodels`-dependent tests are still being skipped:
https://travis-ci.org/choderalab/pymbar/jobs/47017281#L522-L523

Let me look into why this might be happening.  It's likely we have to fix this in a separate PR.

smcantab:
ok

kyleabeauchamp:
Looks good to me.  I ran the tests locally as well.

jchodera:
Thanks again!

@kyleabeauchamp : Can we add @smcantab to the list of contributors?

kyleabeauchamp:
Done (edited directly, no PR).

smcantab:
thanks! happy to help

PR #161, opened 2015-01-15
Moved examples from pymbar-examples

kyleabeauchamp:
I basically just copied the `pymbar/examples/` to `pymbar/examples/harmonic-oscillators/` and then copied in the other files from `pymbar-examples`.


mrshirts:
Looks good to me (from a scan)

PR #165, opened 2015-01-19
Avoid setting the global PRNG state

rmcgibbo:
Just noticed this while browsing the code. Better to use `RandomState` than rely on the mutable global state in the default numpy PRNG in library code.


kyleabeauchamp:
Good catch, thanks

mrshirts:
Looks good.

Issue #168, opened 2015-01-26
detect_equilibration_binary_search is not memory safe


smcantab:
today my cluster administrator has shut down 2 nodes because my jobs using detect_equilibration_binary_search had filled 100% of the memory and were swapping dangerously with the disk. I spent the afternoon looking into it and I have narrowed down the problem to the implementation of np.fft which is called by statsmodel through sm.tsa.stattools.acf. 

Turns out that "fft stores a cache of working memory for different sizes of fft's, so you can run into memory problems if you call this too many times with too many different n's", which is exactly what we do during the binary search. 

I have added my 2 cents to [an issue](https://github.com/numpy/numpy/issues/1864#issuecomment-71480635) that I found in the numpy github repo. My solution at the moment is to add a call to 
```python
np.fft.fftpack._fft_cache.clear()
```
which clears the global variable ```_fft_cache``` after each for loop or after each iteration. Would you be interested in a PR that addresses this issue?

kyleabeauchamp:
I think that sounds like a reasonable workaround.

smcantab:
it is necessary to clear the cache after every fft call, so I have added the necessary lines to the statisticalInefficiency_fft function in PR #169

smcantab:
P.S. tests pass locally

kyleabeauchamp:
Thanks!

smcantab:
Thanks for merging!

PR #169, opened 2015-01-26
make fft calls memory safe

smcantab:
this follows issue #168


js850:
You might want to make that optional, i.e. add a keyword.  I can imagine situations where you want to repeatedly call `statisticalInefficiency_fft` with arrays of the same size.

Also, it's worth noting that this depends on the particular implementation of `sm.tsa.stattools.acf`.  If they start using the scipy version of fft, then clearing `np.fft.fftpack._fft_cache` will not help.  Maybe you should add a comment making that point.

smcantab:
@js850 thanks, I have added a "memsafe" kwarg + documentation and a comment

jchodera:
Thanks!  

Shouldn't we have `memsafe` set to `True` by default, so the default behavior is potentially slower but safer?

smcantab:
I went for the safe option first, then changed my mind. I'll change it again.

smcantab:
done

PR #171, opened 2015-02-18
[WIP] Faster Expectation Calculation

kyleabeauchamp:
Added expectation performance fix and timing code, as well as helper functions for testsystem generation.

For the master branch, we see this:

```
In [23]: %time A, dA = mbar.computeExpectations(x, compute_uncertainty=False)
CPU times: user 29.8 s, sys: 4.65 s, total: 34.5 s
Wall time: 9.65 s

In [24]: %time A = mbar0.computeExpectations(x, compute_uncertainty=False)
CPU times: user 111 ms, sys: 3.4 ms, total: 115 ms
Wall time: 114 ms
```

The current pull request reduces this to:

```
In [23]: %time A, dA = mbar.computeExpectations(x, compute_uncertainty=False)
CPU times: user 515 ms, sys: 70.8 ms, total: 586 ms
Wall time: 342 ms

In [24]: %time A = mbar0.computeExpectations(x, compute_uncertainty=False)
CPU times: user 111 ms, sys: 3.72 ms, total: 115 ms
Wall time: 114 ms
```

### DIFF
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -620,7 +620,7 @@ class MBAR:
         N_k[0:K] = self.N_k
         f_k[0:K] = self.f_k
 
-        # Pre-calculate some quantities outside the for loops
+        # Pre-calculate the log denominator: Eqns 13, 14 in MBAR paper
         states_with_samples = (self.N_k > 0)
         log_denominator_n = logsumexp(self.f_k[states_with_samples] - self.u_kn[states_with_samples].T, b=self.N_k[states_with_samples], axis=1)
         # Compute row of W_nk matrix for the extra states corresponding to u_ln 
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -620,13 +620,18 @@ class MBAR:
         N_k[0:K] = self.N_k
         f_k[0:K] = self.f_k
 
+        # Pre-calculate some quantities outside the for loops
+        states_with_samples = (self.N_k > 0)
+        log_denominator_n = logsumexp(self.f_k[states_with_samples] - self.u_kn[states_with_samples].T, b=self.N_k[states_with_samples], axis=1)
         # Compute row of W_nk matrix for the extra states corresponding to u_ln 
         # that the state list specifies
         for l in L_list:
             la = K+l  #l, augmented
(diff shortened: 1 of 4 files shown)


pymbar/mbar.py
```diff
@@ -620,13 +620,18 @@ def computeExpectationsInner(self, A_n, u_ln, state_map,
...
         # that the state list specifies
         for l in L_list:
             la = K+l  #l, augmented
-            Log_W_nk[:, la] = self._computeUnnormalizedLogWeights(u_ln[l,:]) 
-            f_k[la] = -logsumexp(Log_W_nk[:, la])
-            Log_W_nk[:, la] += f_k[la]
+            # Calculate log normalizing constants and log weights via Eqns 13, 14
+            log_C_a = -logsumexp(-u_ln[l] - log_denominator_n)
+            Q = log_C_a - u_ln[l] - log_denominator_n
+            Log_W_nk[:, la] = Q
```
mrshirts: Maybe eliminate the intermediate Q?  I think it might obscure the code a bit.

pymbar/mbar.py
```diff
@@ -620,13 +620,18 @@ def computeExpectationsInner(self, A_n, u_ln, state_map,
         N_k[0:K] = self.N_k
         f_k[0:K] = self.f_k
 
+        # Pre-calculate some quantities outside the for loops
```
mrshirts: Can you be more specific what "some quantities" are?  It would useful for people learning the methods exactly what is calculated where at the equation level is.

kyleabeauchamp:
OK, made those two fixes

mrshirts:
This is great!  Thanks for recognizing the inefficiency.  The good thing about a single inner loop setup is that it only needs to be fixed in one place.

kyleabeauchamp:
We're still not quite matching the performance of MBAR 2.0, but it's within a factor of 3, which is good enough for expectations, which aren't rate limiting in most cases anyway.

mrshirts:
Sounds good - we can continue to make tweaks as we go.  Indeed,
expectations should not be time limiting.  I can look over the timings and
help tweak as the paper is prepared.

PR #178, opened 2015-04-02
Changed MBAR.computeOverlap to return only O, not the entire tuple. Addi...

wesbarnett:
...tionally

removed argument which was causing the following error:

File "/opt/alchemical-free-energy/alchemical_analysis.py", line 245, in
estimatewithMBAR
    O = MBAR.computeOverlap('matrix')
TypeError: computeOverlap() takes exactly 1 argument (2 given)


kyleabeauchamp:
Thanks, looks good to me.  I will merge once Travis finishes tests.

Issue #185, opened 2015-05-23
Overflow in BAR when using 1-1-48 soft-core in solvation simulations


wesbarnett:
The pymbar implementation of BAR complains "The input data results in overflow in BAR" for solvation simulations when using a [soft-core potential of the 1-1-48 form](http://dx.doi.org/10.1063/1.3697833), as implemented in GROMACS. The GROMACS analysis tool "gmx bar" doesn't have this problem and is able to calculate the results.

In particular I simulated TraPPe-UA methane and ethane in TIP4P/2005 water at 0.1 vdw-lambda increments (from turned off to completely turned on at lambda=1) using the 1-1-48 soft-core potential. MBAR itself gives a result that is consistent with what is expected (as do other methods); however BAR conks out with overflow errors at states 9 -- 10 in both methane and ethane cases and just outputs zero. When using the standard soft-core potential in GROMACS, pymbar's BAR does fine. This overflow problem only happens with the 1-1-48 soft-core potential with free energy of solvation simulations. A simulation of transforming a group to another using the 1-1-48 soft core potential does not result in BAR overflow errors. Again, GROMAC's "gmx bar" handles the results here fine.

jchodera:
Should be simple to fix! Can you post some problematic test data?

wesbarnett:
Great! [Here's the methane data.](https://wavetulane-my.sharepoint.com/personal/jbarnet4_tulane_edu/_layouts/15/guestaccess.aspx?guestaccesstoken=kIvO4%2bUhoWlLCNRKW3k3S9%2finxFv%2bBt7Vyufh8TLmhw%3d&docid=046010034ef1b4c9ba4631ccec1ecc65d)

mrshirts:
Wes, are you using the alchemical-gromacs.py driver to do this, or did you write your own driver?

wesbarnett:
@mrshirts I used the alchemical-analysis.py script.

mrshirts:
Great, that will make it easy to replicate any issues.

jchodera:
Here is the test script I used:
https://gist.github.com/jchodera/33624ca0c37c7e2baff4

mrshirts:
OK, I see where the issue is.  I'll need to cogitate a bit about how to fix it.

PR #186, opened 2015-05-23
Minor bugfix to BAR

jchodera:
I'm trying to make some stability improvements to address #185.

The current BAR code seems pretty complicated.  Not quite sure what it is doing anymore, so it may take me a bit of time to slog through and find the issues.

### DIFF
--- a/pymbar/bar.py
+++ b/pymbar/bar.py
@@ -252,7 +252,7 @@ def BAR(w_F, w_R, DeltaF=0.0, compute_uncertainty=True, maximum_iterations=500, 
                 # Convergence is achieved.
                 if verbose:
                     print("Convergence achieved.")
-                relative_change = 10 ^ (-15)
+                relative_change = 10 ** (-15)
                 break
 
         if method == 'bisection':


kyleabeauchamp:
Is this ready to be tested on the problematic dataset?

jchodera:
I have a test script (locally) to test the problematic dataset, but I haven't yet found which part of the code causes the instability.  The code has now grown past my ability to easily comprehend what it does, so this will take some time.

kyleabeauchamp:
I wonder if it makes sense to deprecate the public-facing BAR and tell users to use MBAR instead.  We could keep the BAR code as a reference for unit tests of 2-state MBAR.

jchodera:
I think other people will need the BAR functionality for a variety of purposes:
* it should be far more robust than MBAR
* comparison tests
* lots of codes still don't support multistate energy computation

If we do want to reduce complexity, we should have BAR use a call to MBAR internally, but I don't think we should deprecate the external BAR interface.

kyleabeauchamp:
I'm happy to defer to you on this.  I only mentioned this because I personally haven't touched the BAR code and you also seem to find parts of it unfamiliar.

jchodera:
Does anyone know who wrote the current BAR code, and what the various weird options are?  Was this @mrshirts?

I'm tempted to eliminate these options to maximize robustness and our ability to support the code.

mrshirts:
I originally wrote it a while ago.  I did a lot of things to make it very
fast and to catch a lot of weird errors that people would notice -- BAR
tends to fail in weird ways when the distributions do not have any real
overlap.  I will take some time tonight to comment things.

For example, one speedup was to use false-position (zeroing a function)
instead of self-consistent iteration, which is converges usually in 5-6
iterations to high precision, instead of many, and is a closed method, so
doesn't diverge (as long as no overflow). I also have bisection in there
which  was used to test; in theory, that could be removed (although maybe I
left it in for some reason?)

The "iterated solution" option is so that one can use the "unoptimized"
BAR, which is the transition matrix estimator for the free energies.  It
just does a single evaluation.

jchodera:
More than commenting is needed---I think this BAR code needs to be massively restructured if we want to keep it.

Can you justify keeping any specific features?  If not, I think we should rip it out and just have it call MBAR internally.

mrshirts:
>
>
> Can you justify keeping any specific features? If not, I think we should
> rip it out and just have it call MBAR internally.
>
Well, I just stated why false position vs. self-consistent iteration --
better convergence and speed.  Do we need it?  It was useful in the past,
don't know if it is now. Could it be that bisection is fast enough?
Possibly.

I also stated why the 'iteration' didn't work.

Why don't I look at the data set, and you not worry about the BAR code?
There are other things you can be doing right now that you'll get more
value out of.

> More than commenting is needed---I think this BAR code needs to be
massively restructured if we want to keep it.

If we can solve this particular convergence issue, then this seems like a
low yield investment -- there is 'needs' and 'would be nice'. Writing the
MBAR paper would be better.

jchodera:
If you want to keep the contents, I'll cede the fixing of #185 to you, @mrshirts.

Merging this change (which doesn't fix #185, but fixes another problem I spotted).

davidlmobley:
Just to weigh in on this - I don't currently have examples on hand, but we
have occasionally run into instances where we get something different out
of BAR and MBAR and in my view this alone is enough of an argument to NOT
just have BAR call MBAR. (My recollection is that we've seen it in both
directions, so sometimes MBAR giving something "better" and sometimes BAR
giving something "better"). Obviously these type of cases ought to be sort
of extreme examples/pathological in some sense, but...

On Wed, May 27, 2015 at 9:10 AM, John Chodera <notifications@github.com>
wrote:

> I think other people will need the BAR functionality for a variety of
> purposes:
>
>    - it should be far more robust than MBAR
>    - comparison tests
>    - lots of codes still don't support multistate energy computation
>
> If we do want to reduce complexity, we should have BAR use a call to MBAR
> internally, but I don't think we should deprecate the external BAR
> interface.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/choderalab/pymbar/pull/186#issuecomment-105980368>.
>



-- 
David Mobley
Associate Professor
Department of Pharmaceutical Sciences
Department of Chemistry
3134B Natural Sciences I
University of California, Irvine
Irvine, CA 92697
dmobley@uci.edu
work (949) 824-6383
cell (949) 385-2436

jchodera:
Please let us know when you see differences between MBAR and BAR, @davidlmobley!  We can't fix issues if we don't have examples...

davidlmobley:
I'll certainly try to if I run across it again. Though to clarify - the
case I was talking about is actually NOT quite a case where a difference is
a problem. I'm talking about a case where I feed *additional* data into
MBAR (evaluations at additional lambda values other than just neighbors)
whereas I only put the pairwise data into BAR. So it's not clear that the
two ought to agree, since the data going in is different. So probably not
as clear cut as I originally made it sound.

On Wed, May 27, 2015 at 10:38 AM, John Chodera <notifications@github.com>
wrote:

> Please let us know when you see differences between MBAR and BAR,
> @davidlmobley <https://github.com/davidlmobley>! We can't fix issues if
> we don't have examples...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/choderalab/pymbar/pull/186#issuecomment-106007520>.
>



-- 
David Mobley
dmobley@gmail.com
949-385-2436

jchodera:
>  I'm talking about a case where I feed *additional* data into MBAR (evaluations at additional lambda values other than just neighbors) whereas I only put the pairwise data into BAR.

Ah, OK.  Nevermind, then---there's no reason these should necessarily agree, especially for poorly sampled systems.  I thought you ha meant you fed the same two-state data into both and got significantly different results.

mrshirts:
MBAR -for just 2 states- should be exactly the same thing as BAR.  If it's
not, then we need to fix this.  MBAR for more than 2 states could of course
be

I think John was referring to BAR just being a wrapper for MBAR with just
two states, not for not using MBAR at all.

mrshirts:
BTW, this particular change is fine to go in.  The reason it wasn't noticed is that this only gets triggered if the free energy is EXACTLY zero so that a relative error can't be computed, and I think even then the logic works most times (will do a check for relative_error < tolerance, which both 10**-15 and -5 are).

PR #189, opened 2015-05-28
BAR improvements.

mrshirts:
Adjusted the logic for avoiding overflows to be more robust.
Simplified the function that needs to be zeroed in BARzero, and documented it better.

### DIFF
--- a/pymbar/bar.py
+++ b/pymbar/bar.py
@@ -52,10 +52,18 @@ import numpy.linalg
 from pymbar.utils import ParameterError, ConvergenceError, BoundsError, logsumexp
 from pymbar.exp import EXP
 
-
-
 def BARzero(w_F, w_R, DeltaF):
-    """Bennett acceptance ratio function to be zeroed to solve for BAR.
+    """A function that when zeroed is equivalent to the solution of
+    the Bennett acceptance ratio.
+
+    from http://journals.aps.org/prl/pdf/10.1103/PhysRevLett.91.140601
+    D_F = M + w_F - Delta F
+    D_R = M + w_R - Delta F
+
+    we want:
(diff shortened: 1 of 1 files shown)


mrshirts:
Addresses Issue #185, plus added comments and reorganization to make the code easier to read.

wesbarnett:
I checked this out and it does solve the issue. But it does add the following warning:

    /usr/lib/python2.7/site-packages/pymbar-3.0.0.dev0-py2.7-linux-x86_64.egg/pymbar/bar.py:347: RuntimeWarning: overflow encountered in exp
  fF = 1 / (1 + np.exp(w_F + C))

mrshirts:
OK, I see that now. Note that the uncertainties will be correct, since all +infs only show up as denominators.  I'll think tonight about the best way to avoid that error, though.

mrshirts:
OK, check this version -- a bit of overkill avoiding overflows, but any slowdown should be pretty much irrelevant for BAR.

wesbarnett:
Warning is eliminated. No problems with BAR now.

jchodera:
Could we add a nosetest for this dataset or something that would trigger
the same overflow issues?

mrshirts:
I think it would be possible. What we'd need was a sample whose energy in the new configuration overflowed exp. If we wanted to take this dataset, then it could be run with just states 9 and 10.

jchodera:
What about just sampling harmonic oscillator test data and modifying one of the snapshot energy differences to be larger than `np.log(sys.float_info.max)`?

kyleabeauchamp:
Is this good to merge or are we still waiting for more tests?  We should get this done soonish so we can roll out the beta release.

mrshirts:
I would suggest we add a test for this problem as an issue and merge it.
We'd love to spend the time to do the test now, but there are a ton of
other things right now that ALSO need to get done.

Issue #192, opened 2015-06-16
Tests are failing on appveyor


jchodera:
From the https://github.com/omnia-md/conda-recipes repo:

https://ci.appveyor.com/project/rmcgibbo/conda-recipes/build/1.0.80/job/jmpmtykan53cc1ja#L491

kyleabeauchamp:
OK, this is because I'm using the HOME environment variable for locating the optional pymbar datasets files.  There's probably a workaround if we want to support windows.

kyleabeauchamp:
I imagine that the MSMBuilder datasets stuff would encounter the same issue, so presumably we can copy paste whatever solution that have there.

jchodera:
Can you be a bit more specific about which part of MSMBuilder or its tests would retrieve optional datasets to test?

kyleabeauchamp:
I'm not sure how well the MSMB code behaves on windows, actually.  It's here: https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/example_datasets/base.py#L45

It's basically a matter of handling an exception and looking in an appropriate place on windows.

rmcgibbo:
Why wouldn't that work just fine on windows? https://docs.python.org/3/library/os.path.html#os.path.expanduser

rmcgibbo:
(i.e., use os.path.expanduser('~') instead of os.environ['HOME'])

kyleabeauchamp:
Cool.  Sounds like we should just backport that function from MSMB.

jchodera:
Want me to do this?

kyleabeauchamp:
From an earlier email conversation, I think @Lnaden was volunteered by @mrshirts.  However, @Lnaden would likely not complain if you did it.

jchodera:
OK, I think I've got it fixed.  I'll add appveyor testing to be sure.

jchodera:
This is all fixed now, and appveyor tests are working.

Should we cut a new release and update the (currently failing) pymbar recipe in https://github.com/omnia-md/conda-recipes ?

kyleabeauchamp:
Sure, might be good to see if we can get windows builds online.

davidlmobley:
Re Windows - just one vote for yes. From a teaching perspective it is very helpful to have things working in Windows. About half of my class this spring had Windows and fortunately everything we needed worked.



—

PR #193, opened 2015-06-16
Replaced hardcoded data home directory with MSMBuilder get_data_home()

jchodera:
This addresses #192 where tests were failing on appveyor.


jchodera:
Local nosetests pass.

kyleabeauchamp:
LGTM

PR #194, opened 2015-06-16
Added appveyor integration based on MSMBuilder

jchodera:
Adding appveyor integration based on MSMBuilder


jchodera:
Looks like the failures were just 32-bit environments, where our tests ran them out of memory.

Issue #199, opened 2015-07-06
comparison to `None` will result in an elementwise object comparison


mrshirts:
We should probably address the warning: 

"FutureWarning: comparison to `None` will result in an elementwise object comparison in the future.
  if N_k == None::"

Sooner rather than later so that people don't worry.  Levi, can you manage that? Also, Levi, not sure why you don't appear in the possible assignee lists.  Perhaps I need to add you . . .

jchodera:
Added @Lnaden to maintainers and developers list.

jchodera:
Be sure to make a PR---don't commit to this repo directly!

Lnaden:
There are several comparisons that will incur this error, specifically the first instance of `kwarg == None` where `kwarg`'s type is a numpy array. I know this is to set the default value of the `kwarg` in a more complex way than the function's `def` statement will allow.

There are 2 ways I can fix this, depending on what we expect the user to feed in.
1) Change all the `== None` statements that can be arrays to `is None`. If any array is set to the `kwarg`, this test returns `False` without tripping the warning.
2) Test for both the `None` object, an empty array, or an array of `None` objects. 

The first case I can come up with a PR to address the instances that will trip the warning pretty quick. The second case will also check if the user passes in an empty array or an array of `None`, but I don't have a clean solution that does not trip the warning yet. 

 I think the first case would be better since the treatment would be the same to an empty array or `None` array would be the same, unless we want to handle those two arrays differently.

Lnaden:
PR #201 should resolve this. I did not go through the `old_mbar.py` file and make changes since I figured people will not care about a future warning on old files.

PR #201, opened 2015-07-08
Future-proof NumPy ndarray == None and ndarray != None comparisons

Lnaden:
This PR resolves issues #199 to prevent NumPy throwing a future warning about comparisons to `None` becoming element wise in future versions.

The PR makes the following replacements to preserve the logic statements functionality in the future: 
`ndarray == None` to `ndarray is None` 
`ndarray != None` to `ndarray is not None`


jchodera:
Thanks!

Issue #198, opened 2015-07-06
Issues with odd input data?


jchodera:
I have to dump out the data I'm feeding `pymbar` to see what is causing these issues, but I'm getting some odd output from `pymbar3`:
```
/Users/choderaj/anaconda/lib/python2.7/site-packages/pymbar-3.0.0.dev0-py2.7-macosx-10.5-x86_64.egg/pymbar/utils.py:70: FutureWarning: comparison to `None` will result in an elementwise object comparison in the future.
  if N_k == None:
/Users/choderaj/anaconda/lib/python2.7/site-packages/scipy/optimize/minpack.py:237: RuntimeWarning: The iteration is not making good progress, as measured by the 
  improvement from the last ten iterations.
  warnings.warn(msg, RuntimeWarning)
/Users/choderaj/anaconda/lib/python2.7/site-packages/pymbar-3.0.0.dev0-py2.7-macosx-10.5-x86_64.egg/pymbar/mbar.py:240: FutureWarning: comparison to `None` will result in an elementwise object comparison in the future.
  if initial_f_k != None:
/Users/choderaj/anaconda/lib/python2.7/site-packages/scipy/optimize/minpack.py:237: RuntimeWarning: xtol=0.000000 is too small, no further improvement in the approximate
  solution is possible.
  warnings.warn(msg, RuntimeWarning)
```

Lnaden:
I get these warnings too with some of my data.

The runtime warnings appear to originate from the `scipy.optimize.root` call on line 321 of `mbar_solvers.py` when the `method = 'hybr'`. Which runtime warning is thrown seems to be a function of the number of samples, whether or not the `initial_f_k`is zeros or the user feeds in the final `f_k` as the `initial_f_k` saved from a previous run, and lastly how much sub-sampling is used. However, I can't find any pattern to determine which one is generated. 

The warnings each indicate that the function may not be converging, however, the final answer may not be wrong if `scipy.optimize.minimize` converges first from the sub-sampling hot-start.

jchodera:
Let's definitely add some tests where multiple schemes are compared to ensure the same answer is obtained, or an analytical system is used to induce these errors.

Is there any way we can avoid these warnings?

Lnaden:
It looks like we can filter out warnings with the `warnings` module based on [this stackoverflow](http://stackoverflow.com/questions/13057022/python-scipy-optimize-using-fsolve-with-multiple-first-guesses) question

```language=python
import warnings
[...]

PR #202, opened 2015-07-09
Ensure free energy is moderately accurate, protocol independent

Lnaden:
This PR is to help diagnose #198 as it generates both numerical errors using `scipy.optimize.root` with `method='hybr'` as well as assert that the free energy is accurate independent of the protocol used to solve MBAR. The assert uses the same basic logic as the `test_mbar.py:test_mbar_free_energies()` function, but cycles through most solver combinations.

One possible change is to include `scipy.optimize.root` methods in the subsampling protocol, and `scipy.optimize.minimize` methods in the solver protocol to be more thorough.

### DIFF
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -42,7 +42,7 @@ import math
 import numpy as np
 import numpy.linalg as linalg
 from pymbar import mbar_solvers
-from pymbar.utils import kln_to_kn, kn_to_n, ParameterError, logsumexp
+from pymbar.utils import kln_to_kn, kn_to_n, ParameterError, logsumexp, check_w_normalized
 
 DEFAULT_SOLVER_PROTOCOL = mbar_solvers.DEFAULT_SOLVER_PROTOCOL
 DEFAULT_SUBSAMPLING_PROTOCOL = mbar_solvers.DEFAULT_SUBSAMPLING_PROTOCOL
@@ -1486,26 +1486,7 @@ class MBAR:
         if(np.sum(N_k) != N):
             raise ParameterError('W must be NxK, where N = sum_k N_k.')
 
-        # Check to make sure the weight matrix W is properly normalized.
-        tolerance = 1.0e-4  # tolerance for checking equality of sums
-
-        column_sums = np.sum(W, axis=0)
-        badcolumns = (np.abs(column_sums - 1) > tolerance)
-        if np.any(badcolumns):
-            which_badcolumns = np.arange(K)[badcolumns]
-            firstbad = which_badcolumns[0]
-            raise ParameterError(
-                'Warning: Should have \sum_n W_nk = 1.  Actual column sum for state %d was %f. %d other columns have similar problems' %
-                (firstbad, column_sums[firstbad], np.sum(badcolumns)))
-
-        row_sums = np.sum(W * N_k, axis=1)
-        badrows = (np.abs(row_sums - 1) > tolerance)
-        if np.any(badrows):
-            which_badrows = np.arange(N)[badrows]
-            firstbad = which_badrows[0]
-            raise ParameterError(
-                'Warning: Should have \sum_k N_k W_nk = 1.  Actual row sum for sample %d was %f. %d other rows have similar problems' %
-                (firstbad, row_sums[firstbad], np.sum(badrows)))
+        check_w_normalized(W, N_k)
 
         # Compute estimate of asymptotic covariance matrix using specified method.
         if method == 'approximate':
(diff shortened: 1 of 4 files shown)


Lnaden:
I have the print for each combination in there as well while we debug #198, we probably want to wait on this PR until we have debugged the problem and I will remove the print

jchodera:
Is this ready to merge?

mrshirts:
Levi, what do you need from us?

Also, what are the changes in the regression tests and examples when this
is run?  Is there a test we can add the previous code failed but this code
catches?

Thanks!

Lnaden:
> what are the changes in the regression tests and examples when this is run?

The `test_protocols()` function I wrote is being checked by the tests on both travis-ci and appveyor and passing, so I am not sure if there are any changes that are needed for the regression tests. 
I originally wrote this test to examine the two `RuntimeWarnings` raised by `numpy` when solving MBAR as per #198, however I extended its functionality to ensure that MBAR is being solved correctly for all solvers. I don't think this should be included in the examples since its more of a diagnostic and robustness test than an example that someone would run to see a specific functionality.

> Is this ready to merge?

 I would not merge this quite yet because it does not guarantee the `RuntimeWarnings` will be raised, dependent on the data. There does not appear to be any intuitive way to raise one `RuntimeWarning` and not the other, however this test often raises at least one of them each time its called. I also have not figured out what to do when we catch the `RuntimeWarning` since warning does not indicate inaccurate free energies (the test would fail otherwise), and the message is only generated after the data has been run through numpy's C code, which `pdb` has issues stepping into. Is there something we should do when the `RuntimeWarning` is raised, or do we want to leave the test as is to ensure that we are at least solving MBAR correctly and debug the `RuntimeWarnings` separately?

Here is the call stack for reference of where the `RuntimeWarnings` are generated, I replaced the protocols with their explicit variables
```python
 File "/pymbar/tests/test_mbar_solvers.py", line 99, in test_protocols
    mbar = pymbar.MBAR(u_kn, N_k, subsampling_protocol=({'method':'L-BFGS-B'},), solver_protocol=({'method':solver_protocol},))
  File "/pymbar/mbar.py", line 264, in __init__
    self.f_k = mbar_solvers.solve_mbar_with_subsampling(self.u_kn, self.N_k, self.f_k, solver_protocol, subsampling_protocol, subsampling, x_kindices=self.x_kindices)
  File "/pymbar/mbar_solvers.py", line 486, in solve_mbar_with_subsampling
    f_k_nonzero, all_results = solve_mbar(u_kn[states_with_samples], N_k[states_with_samples], f_k[states_with_samples], solver_protocol=solver_protocol)
  File "/pymbar/mbar_solvers.py", line 371, in solve_mbar
    f_k_nonzero, results = solve_mbar_once(u_kn_nonzero, N_k_nonzero, f_k_nonzero, **options)
  File "/pymbar/mbar_solvers.py", line 321, in solve_mbar_once
    results = scipy.optimize.root(grad, f_k_nonzero[1:], jac=hess, method=method, tol=tol, options=options)
  File "/scipy/optimize/_root.py", line 176, in root
    sol = _root_hybr(fun, x0, args=args, jac=jac, **options)
  File /scipy/optimize/minpack.py", line 237, in _root_hybr
    warnings.warn(msg, RuntimeWarning)
```
The C code is called at 
```python
 /scipy/optimize/minpack.py(214)_root_hybr()
-> retval = _minpack._hybrj(func, Dfun, x0, args, 1,  
        col_deriv, xtol, maxfev, factor, diag)
```

Lnaden:
I have redone this PR to detect and handle the SciPy `RuntimeWarnings`, check the result, and raise a warning of our own asking the user to verify. The trade off is MBAR displays a bit more when the warning is thrown. I can change it to just detect the warnings and raise an error when MBAR does not converge, if you think that would be a better option.

I also moved the check for `W_nk` normalization to a utility.

I think the appveyor failure is not related to the code and failed from timeout. I can cut the new test's time down by about 20-30% if we need a faster test.

Lnaden:
I have reduced the number of total samples in the solver test I added for speed. The output warning and our response report cleaner as well.

Lnaden:
Anyone had a chance to look over this PR yet?

jchodera:
Sorry for taking so long to get to this!

Can you clarify what is meant by
> I have redone this PR to detect and handle the SciPy RuntimeWarnings, check the result, and raise a warning of our own asking the user to verify. 

Do you mean that we just raise a `RuntimeWarning` with a clearer explanation of the problem?

Lnaden:
Basically, yes. 

I've made it so that if SciPy raises a warning, the code then checks that the solution to `W_nk` is properly normalized to ensure its at least constrained properly. if not quite to the tolerance that SciPy wants. The user is then notified that both a warning was detected, and that `W_nk` is normalized correctly.

Since that alone is not a sufficient check, I added a test to run through the common SciPy solver permutations and check that the free energies are moderatley correct for all of them, similar to the `test_mbar_free_energies` test in `tests/test_mbar.py`.

jchodera:
Sounds great! ok with me to merge

PR #204, opened 2015-07-23
Stop mocking out numpy and scipy in sphinx conf.

kyleabeauchamp:
On my machine, this clears up the sphinx build errors and builds useful API docs again.  Haven't tested on readthedocs, so I suppose the best way to do that is to just merge.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,12 +15,14 @@
 import sys
 import os
 
+# KAB: As of July 22, 2015 it seems that it might be preferable to no longer mock out numpy and scipy.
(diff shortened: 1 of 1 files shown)


jchodera:
There seems to be a numerical failure in py27 travis:
https://travis-ci.org/choderalab/pymbar/jobs/72229057#L546-L561

Any ideas?

kyleabeauchamp:
No idea, but that is a separate issue.

jchodera:
Must have been an odd fluke?

PR #218, opened 2016-03-22
Update pymbar-dev builds with newer devtools framework

jchodera:


jchodera:
Looks like the single test failure was due to an incorrectly coded statistical test.  I'll fix it in a separate PR.

PR #225, opened 2016-05-26
Autodoc now generates class docs from __init__ as well.

dotsdl:
The primary class of ``pymbar``, ``MBAR``, has documentation components
in both the class doc and its ``__init__`` method. The docs in
``__init__`` were ignored by autodoc when generating the sphinx docs,
but now we include them.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -52,6 +52,10 @@ autosummary_generate = True
 autodoc_default_flags = ['members', 'inherited-members']
 numpydoc_show_class_members = False
 
+# concatenate both class and __init__ docstrings when generating autodoc class
+# docs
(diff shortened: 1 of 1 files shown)


jchodera:
@Lnaden Can you take a look at this PR at some point?

Lnaden:
Yeah. Ill put it on my list to handle this week

Lnaden:
@jchodera can we force the appveyor builds to restart in any way? I think the failures are actually scholastic.

jchodera:
Restarted

jchodera:
> scholastic

stochastic?

Lnaden:
Right that one, bad spell check.

PR #230, opened 2016-07-01
Correlation

ChayaSt:
This PR adds the following options to `normalizedFluctuationCorrelationFunctionMultiple`:

1) return D(t) - the unnormalized correlation function `<A(t) B(t)>`
2) Ability to stop accumulating the correlation function once it goes negative.

### DIFF
--- a/pymbar/tests/test_timeseries.py
+++ b/pymbar/tests/test_timeseries.py
@@ -140,3 +140,18 @@ def test_detectEquil_constant_trailing():
     ~50 if we try to include part of the equilibration samples, or it can be Neff=1 if we find that the
     whole first half is discarded. 
     """
+
+def test_correlationFunctionMultiple():
+    """
+    tests the truncate and norm feature
+    """
+    A_t = [testsystems.correlated_timeseries_example(N=10000, tau=10.0) for i in range(10)]
+    corr_norm = timeseries.normalizedFluctuationCorrelationFunctionMultiple(A_kn=A_t)
+    corr = timeseries.normalizedFluctuationCorrelationFunctionMultiple(A_kn=A_t, norm=False)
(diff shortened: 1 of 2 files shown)


jchodera:
Tests are failing, but for unrelated reasons. I'll create an issue.
Can you add tests for your new features, though?

Lnaden:
Tests are now passing. I do have one question though. The PR here will stop the cross-correlation the first time it observes a negative correlation time. You then capture this one negative correlation time and return it as `C_n[-1]`. Is this intended behavior or did you want to capture everything up to but not including the negative correlation time?

Other than that, I think the code looks good. Do we wan to wait until tests exist for this method, or should we handle that in a separate Issue/PR?

jchodera:
> Tests are now passing. I do have one question though. The PR here will stop the cross-correlation the first time it observes a negative correlation time. You then capture this one negative correlation time and return it in theC_n[-1] entry. Is this intended behavior or did you want to capture everything up to but not including the negative correlation time?

This sounds like good behavior, as long as it is documented!

>  Do we wan to wait until tests exist for this method, or should we handle that in a separate Issue/PR?

Please add tests for this method!

ChayaSt:
I'll add tests.

ChayaSt:
> Tests are now passing. I do have one question though. The PR here will stop the cross-correlation the first time it observes a negative correlation time. You then capture this one negative correlation time and return it in theC_n[-1] entry. Is this intended behavior or did you want to capture everything up to but not including the negative correlation time?

For my current purpose, it doesn't really matter if I capture the one negative correlation but I changed it so that it doesn't capture it. 

I added some tests. Let me know if there's anything else I should test for.

Lnaden:
Looks good to me!

jchodera:
Thanks!

PR #233, opened 2016-09-12
Modified .travis.yml to try and fix failing travis tests.

Lnaden:
Attempting to fix the travis failures. Using the changes `.travis.yml` file from [the perses PR](https://github.com/choderalab/yank/commit/134bd8c2180decacf3ff90a4ae3fb1be597e7429) to fix this one. May require a few more pushes to get right.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ script:
   # Add org channel
   - conda config --add channels ${ORGNAME}
   # Add omnia dev channels
-   - conda config --add channels https://conda.anaconda.org/omnia/label/dev
+  - conda config --add channels https://conda.anaconda.org/omnia/label/dev
   # Build recipie
   - conda build devtools/conda-recipe
   # Install locally-built package
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,10 @@ script:
   - conda create --yes -n test python=$python
   # Activate the test environment
   - source activate test
-  # Add Channels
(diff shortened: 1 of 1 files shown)


Lnaden:
All tests are passing. @jchodera can you double check the changes to the .travis.yml file just to make sure I didn't forget or mistype anything before I merge this?

jchodera:
Looks like the `perses` version needs to be updated---I'll take care of that.

Use the [`yank` model instead](https://github.com/choderalab/yank/blob/master/.travis.yml#L26-L45).

The current version you have omits the creation of an environment with the appropriate Python version, so it's not actually testing anything but python 2.7.

Lnaden:
Do we need point at the Omnia dev channel for this with 
`- conda config --add channels https://conda.anaconda.org/omnia/label/dev`
and also have the local install point at the dev version with
`- conda install --yes --use-local ${PACKAGENAME}-dev`? 
Or can we just drop the first line and only use the non-dev version of the `pymbar` install?

jchodera:
We want both.

The dev channel part brings in the latest `openmm` nightly builds.

The second `--use-local` part brings in the `pymbar-dev` version we just built locally.

We should revamp this scheme at some point to have it make more intuitive sense.

Lnaden:
Tests are passing now, if this looks good, I'll merge it.

jchodera:
That should do the trick!

PR #242, opened 2016-12-30
pymbar 3.0.0 release; clean up README and docs

jchodera:
`pymbar 3.0.0dev2` has been out for ages, so it's time to release it.

I'm also taking a stab at cleaning up the `README.md` and docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,10 +88,14 @@ References
 
   Chodera JD, Swope WC, Pitera JW, Seok C, and Dill KA. Use of the weighted histogram analysis method for the analysis of simulated and parallel tempering simulations. J. Chem. Theor. Comput. 3(1):26-41 (2007).  [DOI](http://dx.doi.org/10.1021/ct0502864)
 
+* The automatic equilibration detection method provided in `pymbar.timeseries.detectEquilibration()` is described here:
+
+  Chodera JD. A simple method for automated equilibration detection in molecular simulations. J. Chem. Theor. Comput. 12:1799, 2016.  [DOI](http://dx.doi.org/10.1021/acs.jctc.5b00784)
+
 License
 -------
 
-`pymbar`` is free software and is licensed under the LGPLv2 license.
+`pymbar` is free software and is licensed under the LGPLv2 license.
 
 
 Thanks
--- a/README.md
+++ b/README.md
@@ -44,8 +44,10 @@ To analyze this data, we first initialize the `MBAR` object:
 ```
 Estimating dimensionless free energy differences between the sampled thermodynamic states and their associated uncertainties (standard errors) simply requires a call to `getFreeEnergyDifferences()`:
 ```python
->>> (Deltaf_ij_estimated, dDeltaf_ij_estimated, Theta_ij) = mbar.getFreeEnergyDifferences()
+>>> (Deltaf_ij, dDeltaf_ij, Theta_ij) = mbar.getFreeEnergyDifferences()
 ```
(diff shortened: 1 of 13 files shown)


jchodera:
This should be ready to review.

We can continue to add more detail to the docs after this.

jchodera:
@mrshirts : Should we include any pointers to your alchemical free energy analysis code for gromacs?

README.md
```diff
@@ -6,66 +6,62 @@
...
 ```
+We have the `nsamples` sampled oscillator positions `x_n` (with samples from all states concatenated), [reduced potentials](http://www.alchemistry.org/wiki/Multistate_Bennett_Acceptance_Ratio#Reduced_potential) in the `(nstates,nsamples)` matrix `u_kn`, number of samples per state in the `nsamples` array `N_k`, and indices `s_n` denoting which thermodynamic state each sample was drawn from.
 
-Next, we extract the dimensionless free energy differences and uncertainties:
-
+To analyze this data, we first initialize the `MBAR` object:
+```python
+>>> mbar = MBAR(u_kn, N_k)
+```
+Estimating dimensionless free energy differences between the sampled thermodynamic states and their associated uncertainties (standard errors) simply requires a call to `getFreeEnergyDifferences()`:
```
Lnaden: I would add something like "where the free energy differences are relative to the first state" to explicitly state that index 0 is the ref state.

setup.py
```diff
@@ -14,8 +14,8 @@
 import six
 
 ##########################
-VERSION = "3.0.0.dev0"
-ISRELEASED = False
+VERSION = "3.0.0"
+ISRELEASED = True
```
Lnaden: Don't forget to unset this

README.md
```diff
@@ -93,16 +88,14 @@ References
 
   Chodera JD, Swope WC, Pitera JW, Seok C, and Dill KA. Use of the weighted histogram analysis method for the analysis of simulated and parallel tempering simulations. J. Chem. Theor. Comput. 3(1):26-41 (2007).  [DOI](http://dx.doi.org/10.1021/ct0502864)
 
-
 License
 -------
 
-Pymbar is free software and is licensed under the GPLv2 license.
+`pymbar`` is free software and is licensed under the LGPLv2 license.
```
Lnaden: Extra back tick on the `pymbar`

README.md
```diff
@@ -6,66 +6,62 @@
...
-* NumPy: http://numpy.scipy.org/
-* SciPy: http://www.scipy.org/
-* NumExpr: https://github.com/pydata/numexpr
-* six
-* nose
-* Some optional graphing functionality in the tests requires the matplotlib library: http://matplotlib.sourceforge.net/
+where `EA_k[k]` is the estimated expectation of the mean oscillator position in thermodynamic state `k`.
 
+See the docstring help for these individual methods for more information on exact usage; in Python or IPython, you can view the docstrings with `help()`.  
+Additional examples can be found in [pymbar-examples](http://github.com/choderalab/pymbar-examples/).
```
Lnaden: So... the `pymbar-examples` page lists the following at its header: 

> NOTE: THESE EXAMPLES HAVE BEEN MOVED TO THE MAIN PYMBAR REPOSITORY. FUTURE DEVELOPMENT WILL HAPPEN THERE.

and has not been updated since mid 2015, might not be a good idea to point people at this for now.

jchodera:
Thanks for the suggestions!

Issue #243, opened 2016-12-31
Nose tests cannot find data set files


wesbarnett:
The nose tests seem to expect the data sets from https://github.com/choderalab/pymbar-datasets to be in a directory in my home folder named `pymbar_data`. However, it seems like some of the files are not matching up. Here is my output from those tests (I assume the first two go with these data sets as well):

    pymbar.tests.test_covariance.test_gas ... SKIP: skip
    pymbar.tests.test_covariance.test_8proteins ... SKIP: skip

    pymbar.tests.test_mbar_solvers.test_gas ... SKIP: Cannot load dataset; skipping test.  Try downloading pymbar-datasets GitHub repository and setting PYMBAR-DATASETS environment variable.  Error was '``/home/wes/pymbar_data/gas-properties/gas-properties.h5`` does not exist'
    pymbar.tests.test_mbar_solvers.test_8proteins ... SKIP: Cannot load dataset; skipping test.  Try downloading pymbar-datasets GitHub repository and setting PYMBAR-DATASETS environment variable.  Error was '``/home/wes/pymbar_data/8proteins/8proteins.h5`` does not exist'

I can't find `gas-properties.h5` or `8proteins.h5` in the pymbar-datasets git repo.

jchodera:
Those datasets are in http://github/choderalab/pymbar-datasets

Full integration of these datasets into the testing and timing framework was underway by @kyleabeauchamp before the project was abandoned.

I don't think that there's a way to actually have these large tests complete in time in travis, but I'll give it a try.

jchodera:
I've also added documentation on how to grab the large datasets (to "Getting Started") if you want to test them yourself!

wesbarnett:
I followed that procedure (clone the pymbar-datasets repo to $HOME/pymbar_data) but the `.h5` files are missing. Is there some processing I need to perform on the data sets? Thanks!

jchodera:
Looks like those were never checked in by @kyleabeauchamp, but the scripts to generate the HDF5 files are checked in. Let me try doing that.

jchodera:
We're going to remove those tests until we have someone who is willing to maintain working versions of them. Thanks for pointing this out!

mrshirts:
>  gas-properties.h5 or 8proteins.h5 in the pymbar-datasets git repo.

[...]

PR #247, opened 2016-12-31
Check out pymbar-datasets before nosetests

jchodera:
This addresses #243 for the automated travis tests, though I'm not sure if they will complete in the time allotted by travis.

I'll add some documentation on checking out the large datasets for local testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ install:
   - bash -x devtools/travis-ci/install.sh
   - export PYTHONUNBUFFERED=true
   - export PATH=$HOME/miniconda/bin:$PATH
+  - export PYMBAR_DATA=$HOME/pymbar-datasets
 
 script:
   # Create a test environment
@@ -23,8 +24,11 @@ script:
   - conda config --add channels ${ORGNAME}
   # Add omnia dev channels
   - conda config --add channels https://conda.anaconda.org/omnia/label/dev
-  # Build recipie
+  # Build recipe
   - conda build devtools/conda-recipe
+  # Check out large test datasets
+  - echo "Checking out local copy of pymbar-datasets (will take some time)s"
+  - git clone https://github.com/choderalab/pymbar-datasets/archive/master.zip $PYMBAR_DATASETS
   # Install locally-built package
   - conda install --yes --use-local ${PACKAGENAME}-dev
   - conda install --yes --quiet pip nose nose-timer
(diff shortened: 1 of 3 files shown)


jchodera:
It looks like [`test_mbar.py`](https://github.com/choderalab/pymbar/blob/master/pymbar/tests/test_mbar.py#L81) uses a test criteria that looks to see whether any estimates are more than 3 standard errors from their analytical or numerical estimates. I've changed this to 6 standard errors, but we should further correct this my the number of tests in the suite and the expected number of total commits for the lifetime of the project to not have false positives show up all the time.

jchodera:
We might be able to use [travis caching](https://docs.travis-ci.com/user/caching/) speed up the retrieval of these large datasets.

jchodera:
@mrshirts : Where does all the stuff in http://github.com/choderalab/pymbar-datasets even come from? It's a total and complete mess right now, and the idea of using HDF5 files to store the matrices is pretty much unworkable due to the size of these files (`gas-properties` is over 200 MB, which exceeds the strict github 100MB file size limit). Is there even any value to this?

If nobody is stepping forward to maintain this, I am going to completely purge it.

jchodera:
Closing this in favor of #250.

mrshirts:
They are all examples that people were having trouble with that we eventually fixed, and they contributed their data because we asked them to, so we could have a database of actual use cases for MBAR (as opposed to idealized ones).

1. gas-properties sourcing is in the README - it's a nice 2D example of various physical properties computed as a function of pressure and temperature. It if of value because it's actually how people are using MBAR.   See: For examples of the full application of property prediction, see:
  
"Force field comparison and thermodynamic property calculation of 
supercritical CO2 and CH4 using molecular dynamics simulations" by Cassiano G. Aimoli, 
Edward J. Maginn, and Charlles R.A. Abreu, in Fluid Phase Equilibria, 368, pp 80–90, 2014,
http://dx.doi.org/10.1016/j.fluid.2014.02.001

and 

"Thermodynamic Properties of Supercritical Mixtures of Carbon Dioxide and Methane: 
A Molecular Simulation Study" by Cassiano G. Aimoli,  Edward J. Maginn, and Charlles 
R.A. Abreu, in Journal of Chemical & Engineering Data, 59 (10), pp 3041–3054, 2014, 
http://dx.doi.org/10.1021/je500120v

2. alchemical-large-datasets are the full datasets for solvation of 3-methylindole that we've used for lots of other examples.

3. 8 proteins is an example of calculating heat capacities for associating alpha helices of large proteins that broke MBAR originally, and we needed to fix it.  Looks like a README needs to be added.

4. I have no idea what k69 is. Kyle added it.

PR #250, opened 2016-12-31
Remove abandoned pymbar-datasets code

jchodera:
This PR removes any code associated with the now-abandoned http://github.com/choderalab/pymbar-datasets, as it was untested and broken.

If someone cares enough to make this work, they can restore the code on the conditions that:
* The code works
* The code is tested
* These tests run as part of the nightly tests to ensure it doesn't stop working at some point


mrshirts:
I disagree with throwing out use cases that 1) are part of how other people actually use MBAR for publications and 2) helped us find problems with MBAR.

HOW to best save such examples is a good question that deserves discussion.

jchodera:
This code simply does not work now. The pytables code doesn't even work with the current pytables API. We have to purge it.

If you'd like to add back a working version that is sustainable---that is, included in continuous integration---at some point in the future, you are more than welcome to. But we CANNOT keep broken, untested code in the main development repository.

jchodera:
The code isn't going way---it will be easy to find if you have the time to resurrect it.

mrshirts:
> This code simply does not work now. The pytables code doesn't even work with the current pytables API. We have to purge it.

Sounds like Kyle broke it in the process of trying to update it to run CI.  It worked fine manually before (I.e. read the README, and run).

We need to have a good way of distributing reference examples, since what other people do with MBAR is extremely valuable.

In terms of CI, it should be REALLY simple.  For each example, when the code is updated, does it still give the same answer as it did before? That should be it.

mrshirts:
> We need to have a good way of distributing reference examples, since what other people do with MBAR is extremely valuable.
>> I think these other issues are good things to discuss next week.

Well, it's not really _my_ code.  It's collaborators code that we found very useful to have, because it broke MBAR.

Discussing how to handle these sounds good.  I think the right way to handle this important, because it would be great to archive how other people are using it, without expecting that we have to make all of their work best-practices.  

I agree that fixing these things are a priority.  But we do need to schedule times fix things -- I can't necessarily find the time on new years eve when my brother and his children are here. . .

jchodera:
What we need is an agreed-upon set of standards for collaborative code development. If we stick to that, we'll make sure I don't *have* to be fixing someone else's broken code on New Year's Eve.

jchodera:
We've started to collect some guidelines here:
https://github.com/choderalab/software-development

I'll make sure to update it to prevent this situation from reoccurring.

PR #251, opened 2016-12-31
Fix pip install issues for python 3.x

jchodera:
* Add 'six' to setup.py requirements
* Add travis tests for pip install

cc: https://github.com/alchemistry/alchemlyb/pull/5


jchodera:
Looks like the compiled extension was causing problems for installation via pypi. It wasn't even being built for python 3.x installs anyway, so we will need to overhaul this.

Lnaden:
Is this still a problem with the Py 3.6 PR I just pushed?

mrshirts:
What are the issues with merging this? Do you need me to take a look?

Lnaden:
I'm going to close this for now, I don't think this is needed with some other changes since pip seems to work fine for now.

Issue #246, opened 2016-12-31
Release 3.0.0


jchodera:
We're planning to release 3.0.0 on Fri 6 Jan.

After that, we should evaluate who should become the primary maintainer of `pymbar`.

If it falls to me, my plan is to do a complete reversion back to `pymbar 2.x` solvers (focusing on the adaptive hybrid method for solving the MBAR self-consistent equations) and release pymbar 4.0.0 as soon as possible.

mrshirts:
Let's talk with Naden/Shirts/Chodera (and others who might want to be involved) soon after the 6th to decide.

Lnaden:
3.0 has been out for a while

Issue #241, opened 2016-12-30
Change license to MIT


jchodera:
We've been moving a bunch of our lab projects to the more permissive [MIT License](https://en.wikipedia.org/wiki/MIT_License) to be more fully compatible with OpenMM licensing, and I'd like to do this on Fri 6 Jan unless there are objections.

Lnaden:
Fixe by 3.0.1

Issue #215, opened 2016-02-12
Ready for official 3.0 release?


jchodera:
@mrshirts @Lnaden @kyleabeauchamp @davidlmobley: Are we OK with cutting an official 3.0 release, updating the conda package, and uploading a new version to `pip`?

jchodera:
I can do this if everyone gives the OK.

Lnaden:
I'd like to see if #202 can be merged so the current `SciPy` solvers have a formal test and its warnings are handled a bit more gracefully. I'm okay with a 3.0 release either way though.

mrshirts:
Hi, all-

Let me take a look at Levi's PR over the next day or so, and then let's
release 3.0.

jchodera:
Sounds good!  I can take a look over the weekend too.

Thanks, @Lnaden!

PR #264, opened 2017-05-18
Conda-Forge Migration and 3.0.1 release

Lnaden:
This PR serves 2 purposes:
1. Cut a stable 3.0.1 release
2. Migrate this package to conda-forge. 

I think the 3.0.1 release is nessicary to bring in some bug fixes related to NumPy.

I think this package is an excellent candidate to migrate from omnia to conda forge as it has no omnia dependencies itself. I will personally handle the creation of the conda-forge things once we have a cut release.


Lnaden:
Fixes #215 
Fixes #246
As there were no objections to the license change in #241, I have switched to the MIT License and this Fixes #241.

Lnaden:
This looks like its ready to go. Any objections or concerns before I merge this?

jchodera:
Go for it!

PR #270, opened 2017-07-24
Doc updates for 3.0.3

Lnaden:
Update docstrings to look better on the RTD website.

No changes to code.

@jchodera this should fix the docs you need for your talk. This look good to you? I think you will still have to trigger the Read The Docs build ones I merge


jchodera:
Thanks! This looks good to me.

Issue #272, opened 2017-07-31
Error when installing 3.0.2 from pypi


jbarnoud:
Pymbar cannot be installed from pypi because `pymbar/_pymbar.c` is missing from the tarball.

I try to install pymbar on python 2.7 in a fresh virtualenv:

```bash
mkvirtualenv pymbar-test -p python2.7
pip install numpy six
pip install pymbar
```

Pip tries to install pymbar in version 3.0.2 but fails with the following error message:

```
Failed to build pymbar
Installing collected packages: pymbar
  Running setup.py install for pymbar ... error
    Complete output from command /coarse/jon/Envs/pymbar/bin/python2.7 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-KRHJLH/pymbar/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-K8SheN-record/install-record.txt --single-version-externally-managed --compile --install-headers /coarse/jon/Envs/pymbar/include/site/python2.7/pymbar:
                                      author     Levi N. Naden and Michael R. Shirts and John D. Chodera
                                author_emaillevi.naden@choderalab.org, michael.shirts@virginia.edu, john.chodera@choderalab.org
                                  data_files                                                          []
                                 descriptionPython implementation of the multistate Bennett acceptance ratio (MBAR) method.
                                download_url                         http://github.com/choderalab/pymbar
                                 ext_modules[<setuptools.extension.Extension instance at 0x7fad744a5440>]
                                     license                                                         MIT
                            long_description
        Pymbar (https://simtk.org/home/pymbar) is a library
        that provides tools for optimally combining simulations
        from multiple thermodynamic states using maximum likelihood
        methods to compute free energies (normalization constants)
        and expectation values from all of the samples simultaneously.
    
                                        name                                                      pymbar
                                 package_dir        {'pymbar.tests': 'pymbar/tests', 'pymbar': 'pymbar'}
                                    packages            ['pymbar', 'pymbar.testsystems', 'pymbar.tests']
                                   platforms                            ['Linux', 'Mac OS X', 'Windows']
                                    requires                       ['numpy', 'scipy', 'nose', 'numexpr']
                                         url                         http://github.com/choderalab/pymbar
                                     version                                                       3.0.2
                                    zip_safe                                                       False
    
    running install
    running build
    running build_py
    creating build
[...]

PR #275, opened 2017-09-05
Fix for missing _pymbar.c file from PyPi Tarball

Lnaden:
Fixes a bug in how the `_pymbar.c` file is added to the tarball from setuptools. Partial resolution to #272, will still need to cut a proper 3.0.3.

Conditions to reproduce:
* Tarball compiled from setuptools in Python 3.X
* Install attemped from Python 2.X

This problem did not manifest on Conda because conda builds each tarball in the correct environment.


jchodera:
Thanks!

PR #280, opened 2018-01-11
Adding back in adaptive solving of the MBAR equations, using the previous algorithm.

mrshirts:
Let me know if I should add additional tests.  Speed tests for the examples we have is not noticeable.  I haven't done extensive speed tests for large scale (Levi, perhaps you can remind me how to do those). I assume it will still be slower.  However, it should hopefully be more robust.  I have made it the default algorithm.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,13 +12,7 @@ environment:
       CONDA_PY: "27"
       CONDA_NPY: "112"
 
-    - PYTHON: "C:\\Python34_64"
-      PYTHON_VERSION: "3.4"
-      PYTHON_ARCH: "64"
-      CONDA_PY: "34"
-      CONDA_NPY: "111"
-
-    - PYTHON: "C:\\Python34_64"
+    - PYTHON: "C:\\Python35_64"
       PYTHON_VERSION: "3.5"
       PYTHON_ARCH: "64"
       CONDA_PY: "35"
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -278,7 +278,7 @@ class MBAR:
                 print("f_k = ")
                 print(self.f_k)
 
-        self.f_k = mbar_solvers.solve_mbar_with_subsampling(self.u_kn, self.N_k, self.f_k, solver_protocol, subsampling_protocol, subsampling, x_kindices=self.x_kindices)
+        self.f_k = mbar_solvers.solve_mbar_with_subsampling(self.u_kn, self.N_k, self.f_k, solver_protocol, subsampling_protocol, subsampling, x_kindices=self.x_kindices, verbose=self.verbose)
         self.Log_W_nk = mbar_solvers.mbar_log_W_nk(self.u_kn, self.N_k, self.f_k)
 
         # Print final dimensionless free energies.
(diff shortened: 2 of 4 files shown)


Lnaden:
Huzzah! It may be a couple days before I can get to this, but thank you!

mrshirts:
Hmm. It seems to be failing with

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself, usually with

pymbar.tests.test_covariance.test_100x100_oscillators ... OR with pymbar.tests.test_mbar_solvers.test_200x50_oscillators ...

I'll have to dig into it to find out why it's taking so long, since that seems to be the one possible issue, changing adaptive to the default.

mrshirts:
Ah, OK, I think I figured it out.  Default tolerance is 1 x 10^-15, whereas adaptive is sometimes, depending on the problem only getting down to 1.0 x 10^-15, and can't get any lower. Given machine precision, I'm not sure the default should be 1.0 x 10^-15. At least for adaptive, that's pretty low already.  Perhaps something like 10^-12 (or at least 10^-14)?  Easy to see how a reasonable method could be machine precision away.

mrshirts:
Some of the AppVeyor fails are the same as the other tests, but there's one I don't get and appears unrelated.

ld\pymbar-dev_1515900639756\test_tmp\run_test.py" 
import: 'pymbar'
  File "C:\Python34_64\conda-bld\pymbar-dev_1515900639756\test_tmp\run_test.py", line 2, in <module>
    import pymbar
  File "C:\Python34_64\conda-bld\pymbar-dev_1515900639756\_test_env\lib\site-packages\pymbar\__init__.py", line 31, in <module>
    from pymbar import timeseries, testsystems, confidenceintervals, version
  File "C:\Python34_64\conda-bld\pymbar-dev_1515900639756\_test_env\lib\site-packages\pymbar\confidenceintervals.py", line 24, in <module>
    import scipy.special
  File "C:\Python34_64\conda-bld\pymbar-dev_1515900639756\_test_env\lib\site-packages\scipy\special\__init__.py", line 640, in <module>
    from ._ufuncs import *
ImportError: DLL load failed: The specified module could not be found.

mrshirts:
Also added handling for adaptive verbose output.  A little kludgy.  Should try to either 1) come up with a better way of handling calls to all solvers or make adaptive look more like a scipy solver (though I don't really like how they are set up).

mrshirts:
The DLL load fail is the only one left - only occurs for python 3.4

jchodera:
Can you review @lnaden?

Lnaden:
I can try to get to this on Friday afternoon, but it may be Sat before I can sit down to review this.

I will say that Python 3.4 failing is less important since conda-forge is Py27, 35, and 36 only

mrshirts:
I may try to tweak the interface a bit in the next day or so, so that all of the methods are called more consistently.  The internals I'm not planning on making changes to (unless some reasons are found). I will make those changes before Saturday (and may make some other smaller cleanup changes in other places as well that would be nice to get reviewed).

Lnaden:
Review in, also took a stab at fixing the AppVeyor tests which drop the 3.4 test and fix a variable for the 3.5 one

mrshirts:
Any guesses on the 3.5 only error?

Lnaden:
There we go, this all passed now @mrshirts!

PR #284, opened 2018-01-30
modified _pymbar extension to allow compilation across python versions

### LINKED ISSUE #253: Overhaul compiled C extensions
We need to ditch the old, broken, py 3.x-incompatible `_pymbar.c` compiled extension and use something more portable and compatible.

For now, it is disabled.

cc: #251 

cc: https://github.com/alchemistry/alchemlyb/pull/5

kamran-haider:
Minor issue: `_pymbar` extension isn't compiled under python3.x. The extension is probably dispensable as it is used only in `old_mbar` and the program doesn't crash if the module it isn't available. Nevertheless, probably a good idea to maintain cross-platform compatibility. This PR should fix  #253.

### DIFF
--- a/pymbar/_pymbar.c
+++ b/pymbar/_pymbar.c
@@ -185,8 +185,40 @@ static PyMethodDef _pymbar_methods[] = {
   {NULL, NULL, 0, NULL}
 };
 
-DL_EXPORT(void) init_pymbar(void)
+//--------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jchodera:
Thanks, @kamran-haider!

We'll hopefully be able to review this later this week.

Tagging @mrshirts and @Lnaden.

mrshirts:
Closing because we will not be supporting _pymbar.c in the future

Issue #281, opened 2018-01-12
Difference in BAR and MBAR uncertainties


mrshirts:
Hi, all-

I've identified an interesting difference between BAR and MBAR uncertainty estimates as currently implemented in the code.  If you start directly from Bennett's paper, plugging equation (8) into (7), you eventually get:

        #             <[f(-x)]^2>_0          <[f(x)]^2>_1            1     1                                                  
        #         [------------------]  + [---------------]    -    --- - ---                                                 
        #            n0 <f(-x)>_0^2          n1 <f(x)>_1^2           n0    n1                                           

Where "_i" indicates averages over state 0 or 1,  f(x) is the Fermi function, and x = M+(F1-F0)-(U1-U0), with  M = ln n1/n0. This formula has been rederived a couple of times in different places, and is what is generally used for BAR -- and is what is used in the code currently. Call this method A.

If, however, you take my PRL BAR paper you get an estimator for the variance that can be reduced to:

        #                                           1                                         1     1                         
        #         --------------------------------------------------------------------    -  --- - ---                        
        #            n0 <f(-x)>_0 - n0 <[f(-x)]^2>_0 + n1 <f(x)>_1 + n1 <[f(x)]^2>_1          n0    n1 

Call this method B. This approach also exactly what MBAR gives you (as proven in the MBAR paper).  This the Fisher-information-derived uncertainty obtained from the likelihood.

A and B are not actually the same thing, though in the limit of good sampling, they are both practically indistinguishable from the true variance (obtained by repeating a bunch of times).  I believe they are both "correct" in that they are both asymptotic estimators of the uncertainty, they just have somewhat different asymptotic behavior.

Which one is better?  

* Method A turns out to be a good estimate even if your free energy is wrong, which method B doesn't d So I was using method A since I was using MBAR for other BAR-based estimators (UBAR and RBAR in the Paliwal et al. 2011 paper).  Recall that Bennett showed that his approach produced an estimator of the free energy even if you didn't use the right constant, but that the variance was minimized with the right constant (Which was the free energy + log ratio of number of samples).   So A works regardless of whether the constant is right.

* Method B appears to be slightly closer to the right answer for moderate to small overlap, but the difference is very small.  

* Method B is wrong if the constant is wrong (IIRC - this is something I call from a while ago, which is why I bothered to change it from the original method B to this one; maybe there was another error that was really the problem).

* Method A is somewhat of an underestimate for very poor overlap and low numbers of samples.  

[...]

PR #286, opened 2018-02-22
BAR uncertainty clarified

mrshirts:
Explained both BAR and MBAR uncertainty estimates, and made it possible to switch to which method. Still not clear which is best, but all the necessary information is there now.
    
Also fixed printing for python3.

### DIFF
--- a/devtools/travis-ci/install.sh
+++ b/devtools/travis-ci/install.sh
@@ -7,9 +7,16 @@ MINICONDA=Miniconda2-latest-Linux-x86_64.sh
 MINICONDA_HOME=$HOME/miniconda
 MINICONDA_MD5=$(curl -s https://repo.continuum.io/miniconda/ | grep -A3 $MINICONDA | sed -n '4p' | sed -n 's/ *<td>\(.*\)<\/td> */\1/p')
 wget -q http://repo.continuum.io/miniconda/$MINICONDA
+# Parse version out of file as fall back
+# Sed cuts out everything before the first digit, then traps the first digit and everything after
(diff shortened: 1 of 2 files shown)


mrshirts:
Addresses #281

mrshirts:
Travis-ci bug was a miniconda mismatch?

+[[ da2dd466d26e33a2b1f72fdb853a8ff0 != dd54b344661560b861f86cc5ccff044b ]]
+echo 'Miniconda MD5 mismatch'
Miniconda MD5 mismatch
+exit 1

Lnaden:
> Travis-ci bug was a miniconda mismatch?

Ugh, this again. This happens when the hosting site pushes a new version (4.4.10 in this case), which updates the links to the `latest` file that the script downloads, BUT they don't update the MD5's on the site yet. You can see that in this case the MD5 its checking from the file (right hand side) matches the 4.4.10 file. Last time this happened I just had to wait a couple days for them to fix it.

One option we could do is parse the version inside the `latest` file that we download, then do a sed parse of the site and check it against the reported MD5 of that specific version on the site. It may be more robust, but also seems like something we shouldn't have to do.

Here is the hosting site:
https://repo.continuum.io/miniconda/

mrshirts:
Are y'all OK with merging this for now since it passes the appveyor tests? Is coverage different? I want to try to do some building on the code (adding more tests, changing the output format).

Lnaden:
I have a workaround fix for the MD5 problem, let me get it in as part of this PR and if the tests pass then we can merge it. I should be able to have it in a few minutes

Lnaden:
I went ahead and merged this, it looked good and the MD5 fix worked.

PR #291, opened 2018-02-23
No string raises

mrshirts:
Eliminate cases where strings are raised as exceptions.

### DIFF
--- a/pymbar/confidenceintervals.py
+++ b/pymbar/confidenceintervals.py
@@ -223,7 +223,7 @@ def generateConfidenceIntervals(replicates, K):
                     print(replicate['error'])
                     print("destimated")
                     print(replicate['destimated'])
-                    raise "isnan"
+                    raise DataError("isnan")
                 else:
                     if abs(replicate['error']) <= alpha * replicate['destimated']:
                         a += 1.0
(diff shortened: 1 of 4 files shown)


pymbar/confidenceintervals.py
```diff
@@ -223,7 +223,7 @@ def generateConfidenceIntervals(replicates, K):
                     print(replicate['error'])
                     print("destimated")
                     print(replicate['destimated'])
-                    raise "isnan"
+                    raise DataError("isnan")
```
Lnaden: This module does not seem to import DataError from anywhere.

mrshirts:
Ah, missed that. Interesting that it passed in 3.x.  I'd like to not make this depend on the rest of pymbar, since its more general and doesn't depend currently.  I'm looking through https://docs.python.org/3/library/exceptions.html#Exception to see if there's a good way to do this but have to come back to this later today.

Lnaden:
I don't think there is any test which touches that part of the code, so there is nothing to test. The only place that function is invoked is in `harmonic-oscillators-distribution.py`, which itself is never invoked anywhere from what I can tell.

mrshirts:
Not sure how it broke this time when the appveyor ones passed.

mrshirts:
OK, since the tests pass, and I've address the comments, I am going to merge myself.  I THINK the tests failed because it uses a random sample, and it's possible for that to fail. Should consider using a fixed seed (now possible with latest check in).

PR #293, opened 2018-02-24
[WIP] Return values a dictionary

mrshirts:
Functions that return multiple values are now all configured to return their values as dictionaries.  This also requires changing tests, examples, and documentation.

First pass, just let me know if you see any issues with how this is set up.  Don't worry about all the details here: I'll fix any of those once we are on the same page.

Also added some corner case testing to the 'adaptive' method in mbar_solvers to fix a test that gave a warning. 

Note, that this change will break backwards compatibility with existing scripts; anything calling pymbar will need to change at least a couple of lines. This will have to be heavily documented in the release notes.


pymbar/bar.py
```diff
@@ -289,10 +297,8 @@ def BAR(w_F, w_R, DeltaF=0.0, compute_uncertainty=True, uncertainty_method='BAR'
             # The free energy difference appears to be zero -- return.
             if verbose:
                 print('The free energy difference appears to be zero.')
-            if compute_uncertainty:
-                return [0.0, 0.0]
-            else:
-                return 0.0
+            result_vals['Delta_f'] = 0.0 
+            break
```
Lnaden: Did you mean to change this to a `break` instead of a return?

pymbar/mbar.py
```diff
@@ -1292,10 +1345,14 @@ def computePMF(self, u_n, bin_n, nbins, uncertainties='from-lowest', pmf_referen
...
         -------
-        f_i : np.ndarray, float, shape=(K)
-            f_i[i] is the dimensionless free energy of state i, relative to the state of lowest free energy
-        df_i : np.ndarray, float, shape=(K)
-            df_i[i] is the uncertainty in the difference of f_i with respect to the state of lowest free energy
+        result_vals : dictionary
+
+        Possible keys in the result_vals dictionary:
+
+        'f_i' : np.ndarray, float, shape=(K)
```
Lnaden: Is there a good reason this is `f_i` and `df_i` instead of `Delta_f` and `dDelta_f` like in `getFreeEnergyDifferences`?

mrshirts: For a PMF, you are required to set an overall reference.  So it's IS a dF from the reference, but not from a specific state.

pymbar/bar.py
mrshirts: No, should be break, so that it can then compute the uncertainty.  But actually, it shouldn't set results there, it should just break, since DeltaF gets copied to results_vals anyway.

PR #295, opened 2018-02-26
Remove Subsampling from MBAR Solving

Lnaden:
As it says in the title. Also cleans up some more warning messages and does better handling of specific ones in testing.


mrshirts:
What do you think about the idea having the option starting with BAR from i->i+1 to initialize? If they are 1D (which is common), that's often an excellent place to start.  

Or we leave adding that to later, and just keep it zeros for now.

mrshirts:
I'll let you resolve the conflicts before I review.

pymbar/mbar.py
```diff
@@ -114,12 +114,12 @@ def __init__(self, u_kn, N_k, maximum_iterations=10000, relative_tolerance=1.0e-
         initial_f_k : np.ndarray, float, shape=(K), optional
             Set to the initial dimensionless free energies to use as a
             guess (default None, which sets all f_k = 0)
-        method : list(dict), optional, default=None
+        solver_protocol : list(dict) or None, optional, default=None
             List of dictionaries to define a sequence of solver algorithms
             and options used to estimate the dimensionless free energies.
             See `pymbar.mbar_solvers.solve_mbar()` for details.  If None,
             use the developers best guess at an appropriate algorithm.
```
mrshirts: Do you want to be more specific about "developers' best guess? Is it just automatically adaptive?

pymbar/mbar.py
```diff
@@ -114,12 +114,12 @@ def __init__(self, u_kn, N_k, maximum_iterations=10000, relative_tolerance=1.0e-
...
-        initialize : string, optional
+        initialize : 'zeros' or 'BAR', optional
```
mrshirts: is the 'try BAR pairwise code' in there right now?

pymbar/mbar.py
```diff
@@ -278,25 +278,17 @@ def __init__(self, u_kn, N_k, maximum_iterations=10000, relative_tolerance=1.0e-
...
-
+        if solver_protocol is None:
+            solver_protocol = ({'method': None},)
         for solver in solver_protocol:
             if 'options' not in solver:
                 solver['options'] = dict()
-                if self.verbose:
-                    # should add in other ways to get information out of the scipy solvers, not just adaptive,
-                    # which might involve passing in different combinations of options, and passing out other strings.
-                    if solver['method'] == 'adaptive':
```
mrshirts: How is the default set here?

pymbar/mbar_solvers.py
```diff
@@ -497,6 +504,9 @@ def solve_mbar(u_kn_nonzero, N_k_nonzero, f_k_nonzero, solver_protocol=None):
     """
     if solver_protocol is None:
         solver_protocol = DEFAULT_SOLVER_PROTOCOL
+    for protocol in solver_protocol:
+        if protocol['method'] is None:
+            protocol['method'] = DEFAULT_SOLVER_METHOD
```
mrshirts: Might this screw things up by placing inconsistent options for the default method?

pymbar/tests/test_covariance.py
```diff
@@ -33,7 +35,8 @@ def _test(data_generator):
     eq(pymbar.mbar_solvers.self_consistent_update(U, N_k, mbar.f_k), mbar.f_k, decimal=10)
 
     # Test against old MBAR code.
-    mbar0 = pymbar.old_mbar.MBAR(U, N_k)
+    with suppress_warnings():
```
mrshirts: Why is suppress_warnings needed here?

pymbar/tests/test_mbar_solvers.py
```diff
@@ -38,7 +41,8 @@ def _test(data_generator):
     eq(pymbar.mbar_solvers.self_consistent_update(U, N_k, mbar.f_k), mbar.f_k, decimal=10)
 
     # Test against old MBAR code.
-    mbar0 = pymbar.old_mbar.MBAR(U, N_k)
+    with suppress_warnings():
```
mrshirts: Ditto

pymbar/tests/test_mbar_solvers.py
```diff
@@ -47,44 +51,39 @@ def test_100x100_oscillators():
...
 
 def test_protocols():
-    '''Test that free energy is moderatley equal to analytical solution, independent of solver protocols'''
-    #Supress the warnings when jacobian and Hessian information is not used in a specific solver
-    import warnings
-    warnings.filterwarnings('ignore', '.*does not use the jacobian.*')
-    warnings.filterwarnings('ignore', '.*does not use Hessian.*')
-    from pymbar.tests.test_mbar import z_scale_factor # Importing the hacky fix to asert that free energies are moderatley correct
+    '''
+    Test that free energy is moderatley equal to analytical solution, independent of solver protocols
```
mrshirts: moderately

Lnaden:
> starting with BAR from i->i+1 to initialize?

Not sure what you mean by this? As in for BAR `f_k` have it be `[0, 1, 2, 3...]`?

pymbar/mbar.py
Lnaden: I'll add some more info about what "default" is

pymbar/mbar.py
Lnaden: Not sure, I'll double check. Most of this PR was just to blow away the `subsampling`, I did not spend time to fix the anything else.

pymbar/mbar.py
Lnaden: No, I suppress the warnings thrown by not having the `verbose` options in the SciPy solvers, but explicitly only that.

pymbar/mbar_solvers.py
Lnaden: Not sure how, the `solver_protocol` is checked first to see if its `None` and assign the default Protocol, then each `solver_protocol['method]` is checked to see if *that* is `None`, to assign the default Method

pymbar/tests/test_covariance.py
Lnaden: It suppresses some [very specific warnings](https://github.com/choderalab/pymbar/blob/fe5731312db92a7a396b777ee455c1ec382d4c1e/pymbar/utils_for_testing.py#L66-L68) complaining about methods which don't have a Hessian or a Jacobian, but will still solve the correct way. This only comes up in the test suite and if a user ever uses those methods, the warnings print as normal. I could change the function to be called something like `suppress_derivative_warnings_for_test` to be more explicit if ti helps?

pymbar/mbar.py
mrshirts: My comment about BAR above was based on this being there.  I had it in at one point, but Kyle removed it, so can be easily exhumed.

pymbar/mbar.py
mrshirts: Ah, I see.  Missed that logic. SEEMS like it's better to handle it explicitly than potentially mask other warning problems?

pymbar/mbar_solvers.py
mrshirts: OK.

pymbar/tests/test_covariance.py
mrshirts: Ah, OK. that makes sense. Yeah, probably better to be more explicit in the name.

mrshirts:
> 
Not sure what you mean by this? As in for BAR f_k have it be [0, 1, 2, 3...]?

BAR generally only works well if i/i+1 are neighbors for all i.   Some of my subsampling replacement ideas will eventually make it work on other cases, but not for now.

mrshirts:
Any other information needed here from me?

Lnaden:
Just one question I have:

For the BAR initialization, is there any way we can assert states are adjacent with each other? I think we may only be able to assume (and document) that behavior, or provide a way for the user to specify

mrshirts:
Right now, the only way to assume that is by the order they are input.  That's what we did before in the previous code; assume 1-2 , 2-3, 3-4, are adjacent. I'm thinking about alternatives, but that's not there yet.  

Maybe the right thing to do is ignore this right now, just do it from zeros, then come up with the right sparse solver to use to initialize in the next couple of week.

Lnaden:
Sounds good. I'll be able to wrap this up tonight

PR #296, opened 2018-02-26
Replace scipy.misc

mrshirts:
Created to remove deprecated scipy.misc code (and a bunch of warnings from CI).


mrshirts:
Wait until CI comes back to review . . . there's a couple things that could cause problems.

mrshirts:
Ah, looks like 3.4 (or at least, whatever scipy is included in 3.4) doesn't have the code change yet.  Perhaps we leave this code in for now, and wait until we rotate out support for 3.4.

jchodera:
Since we are distributing this under omnia, why not align ourselves with 2.7/3.5/3.6 only right now?

mrshirts:
Well, if we want to make that switch now, I'm OK with that.  Somebody else will probably have to remove the 3.4 tests from Travis CI.  I note the appveyor ones already only support 2.7 + 3.5,3.6.

mrshirts:
So, thoughts?  How do we remove the 3.4 testing from Travis CI?  Is that what we want to do?

Lnaden:
I would say remove it. I was going to put that into mine, but if you want to put it here, you can just remove [this line](https://github.com/choderalab/pymbar/blob/replace_scipymisc/.travis.yml#L35) from the .travis.yml

Lnaden:
Thinking about it though, we may want to trap the error in case not every has (or can) upgrade to Scipy 1.0

mrshirts:
I would leave that up to you! I'm don't have a good sense of what the best thing is. 

I'll leave the handling of this PR to you to resolve this.

PR #299, opened 2018-02-28
Add DOI Badge to the README

Lnaden:
As it says on the tin

The DOI tracking has been enabled for a while, this just adds it to the README/Homepage


jchodera:
Thanks!

PR #302, opened 2018-03-01
Bootstrap uncertainties first pass

mrshirts:
Just putting it up here to show what is going on.  Works for free energies so far, still need to get it working for other sections.  No merging yet until that is working!


mrshirts:
PROBABLY this is being supported through alchemlyb instead of pymbar.  Will have to rethink about how to implement this most generally.

pymbar/mbar.py
```diff
@@ -301,7 +302,26 @@ def __init__(self, u_kn, N_k, maximum_iterations=10000, relative_tolerance=1.0e-
...
 
-        self.f_k = mbar_solvers.solve_mbar_for_all_states(self.u_kn, self.N_k, self.f_k, solver_protocol)
+        np.random.seed(rseed)
+        if rseed != None:
+            self.rstate = np.random.get_state()
+
+        f_k_init = self.f_k.copy()  # we need to pass a copy so we don't overwrite the original 
+        self.f_k = mbar_solvers.solve_mbar_for_all_states(self.u_kn, self.N_k, f_k_init, solver_protocol)
+
+        self.bootstraps = None
```
zhang-ivy, line 312: Should this be `self.nbootstraps`?

zhang-ivy:
@mrshirts @jchodera : I'm finding that bootstrapping using this branch takes a long time, since each call to `solve_mbar_for_all_states()` takes about 1.5 seconds for an input potential energy matrix with with 36 states and 300 samples per state. It takes even longer (~35 seconds) for an input matrix of 36 states and ~9000 samples. 

This means that if I bootstrap 1000 times, it takes 30 minutes (for 300 samples) or 10 hours (for 9000) samples to retrieve the free energy and bootstrapped uncertainty.

Is there a way to speed this up?

pymbar/mbar.py
```diff
@@ -527,12 +548,20 @@ def getFreeEnergyDifferences(self, compute_uncertainty=True, uncertainty_method=
 
         result_vals['Delta_f'] = Deltaf_ij
 
-        if compute_uncertainty or return_theta:
+        if compute_uncertainty and uncertainty_method == 'bootstrap':
+            diffm = np.zeros([self.K,self.K,self.nbootstraps])
+            # take the matrix of differences of f_ij
+            for b in range(self.nbootstraps):
+                diffm[:,:,b] = np.matrix(self.f_k_boots[b,:]) - np.matrix(self.f_k_boots[b,:]).transpose()
```
zhang-ivy, line 555: @mrshirts : Could you clarify why we need to subtract each bootstrapped free energy matrix from its transpose?

zhang-ivy:
> Works for free energies so far, still need to get it working for other sections. No merging yet until that is working!

@mrshirts : What do you think of merging this PR and opening another one to deal with the other parts? Is there a test I can help write that would make us more comfortable merging this?

I'm asking because I'd like to use this bootstrapping feature, but the rest of the codebase has changed quite a bit.

mrshirts:
Really good question.  This paused because we were trying to figure out if it made more sense to put it in alchemlyb, since if one is doing analysis of replica exchange or other multiple sets of data at a time, then the bootstrapping should be done at the layer above pymbar, not in pymbar itself.  But we didn't quite resolve that.

zhang-ivy:
>This paused because we were trying to figure out if it made more sense to put it in alchemlyb, since if one is doing analysis of replica exchange or other multiple sets of data at a time, then the bootstrapping should be done at the layer above pymbar, not in pymbar itself.

What do you think of making bootstrapping an option in `pymbar` as well as `alchemlyb`? This would allow the user to choose at which layer they would like to perform the bootstrapping. I use `openmmtools` to run my repex simulations and tried bootstrapping at the openmmtools level, but its quite slow, so I was hoping to do bootstrapping at the pymbar layer. 

tagging @jchodera for his input here as well

jchodera:
@mrshirts : Even if `alchemlyb` adds bootstrapping, that's not really an argument _not_ to also add this optional variance estimation mode that uses bootstrapping to MBAR. MBAR is used in many contexts outside of `alchemlyb`, and we are finding bootstrapping is a useful means of assessing variance in this context.

zhang-ivy:
@mrshirts : Just pinging you again for your feedback on the above discussion!

mrshirts:
@zhang-ivy see #448 for an implementation of this in pymbar4, intended for release soon.  If you can test it out, that would be helpful, still alpha code.

mrshirts:
OK, try this out! One question I have - it looks like this was implemented in a version that already had the API change, as it was a commit to master and not to 3.0.7?  But I guess you were cherry picking that out?

pymbar/mbar.py
```diff
@@ -301,7 +302,33 @@ def __init__(self, u_kn, N_k, maximum_iterations=10000, relative_tolerance=1.0e-
...
+        self.f_k = mbar_solvers.solve_mbar_for_all_states(self.u_kn, self.N_k, f_k_init, solver_protocol)
+
+        self.nbootstraps = None
+        if nbootstraps != None:
+            self.nbootstraps = nbootstraps
+            #save the original data:
+            self.f_k_boots = np.zeros([self.nbootstraps,self.K])
+
+            allN = int(np.sum(self.N_k))
+            rinit = np.zeros(allN, int)
```
zhang-ivy, line 318: @mrshirts : This line should probably be moved to be inside `for b in range(self.nbootstraps):` but outside ` for k in range(K):` right? It's also probably fine where it is.

mrshirts: It should be fine where it is, because all of the states have some value of K, so rinit gets completely refilled each iteration

mrshirts:
> Which API change are you referring to 

3.0.7 returns as tuples, but this commit used a version that returned a dictionary.

zhang-ivy:
Ah I see. When I merged in master, I made sure that my local branch matched output formats in master (by default return a tuple and only return a dict if `return_dict` is True)

mrshirts:
You may want to merge it into the 3.0.7 to make sure its just different by the bootstrapping?

zhang-ivy:
Just double checked how my local branch differs from 3.0.7 and the only difference is that my local bootstrap branch is missing https://github.com/choderalab/pymbar/commit/7ff37e95c92527c6d9e64453ecf892986edb21c6

zhang-ivy:
I'm assuming we'll want to undo the API changes that are present in this branch, so after I merge in https://github.com/choderalab/pymbar/commit/7ff37e95c92527c6d9e64453ecf892986edb21c6, I should push the changes in my local branch to this one?

zhang-ivy:
Ok, I've tested pymbar3 vs pymbar4 free energy estimates with and without bootstrapping and the results are worrisome.

I am using the following openmmtools code (which uses pymbar under the hood) to compute the free energy: 
```python
reporter = MultiStateReporter(path)
analyzer = MultiStateSamplerAnalyzer(reporter,  max_n_iterations=5000)
f_ij, df_ij = analyzer.get_free_energy()
print(f_ij[0,-1], df_ij[0,-1])

```
Output when running with `bootstrap` branch (pymbar3): 
```
-51.57110153691221 0.03839509621355673
```

After subclassing `MultiStateSamplerAnalyzer` functions to use the new pymbar4 API and then running the same code as above, I get the following output when using `pymbar4_bootstrapping` branch:
```
-41.14248325882138 1.6799096188072096
```

Note that the above results do not involve bootstrapping to compute the free energy/uncertainty. 

----

Now for the boostrapping results. Here is the code I am using to bootstrap with pymbar3:
```python
reporter = MultiStateReporter(path)
analyzer = MultiStateSamplerAnalyzer(reporter, max_n_iterations=5000)
decorrelated_u_ln, decorrelated_N_l = analyzer._compute_mbar_decorrelated_energies()
analyzer.mbar = MBAR(decorrelated_u_ln, decorrelated_N_l, nbootstraps=10)
results = analyzer.mbar.getFreeEnergyDifferences(compute_uncertainty=True, uncertainty_method='bootstrap', return_dict=True)
print(results['Delta_f'][0, -1], results['dDelta_f'][0, -1])    
```

Note: I know that 10 bootstrap iterations is not enough, but it takes awhile to run 1000 bootstraps, so I ran with 10 for now just to test that we are getting a larger error compared to without bootstrapping. 

Here are the results when running with pymbar3:
```
-51.57110153691221 0.08023215497413216
```

When I adapt the code above (for computing a bootstrapped free energy) to use the pymbar4 API, I get:
```
-41.14248325882138 7.105427357601002e-15

```

Results:
- I'm seeing that the free energy estimates are very different when comparing pymbar3 vs pymbar4, even without bootstrapping.
- When comparing the uncertainty estimate without vs with bootstrapping, 
    - for pymbar3, the uncertainty increases with bootstrapping, which is what I'd expect.
    - for pymbar4, the uncertainty decreases with bootstrapping, which is worrisome.

Takeaways:
- Something seems very wrong with pymbar4?
- There's probably a much easier way to test this (without using openmmtools) -- I'm happy to implement it if you have a suggestion for the best way to do it.

mrshirts:
I'll take a look at my examples again with pymbar3 and 4 on the plane (about to board). 

I would turn off decorrelation for test cases to make sure there's no difference between versions in terms of the decorrelation (or directly verify that the same number of samples is going into MBAR in both cases). 

Try running stuff in the `examples` directory. 

I would turn on logging as well if possible to see if anything is going wrong with the solver. 

import logging
import sys
logging.basicConfig(stream=sys.stdout, level=logging.INFO)

mrshirts:
In terms of what features to merge - that API message is just for BAROverlap.  I would try to keep it as close to 3.0.7 as possible. Check with @mikemhenry.

mrshirts:
I tried bootstrapping on the harmonic-oscillators example with both pymbar3 and pymbar4, and they gave both very similar standard deviations with and without bootstrapping (see data below).  So maybe an issue with the interface?  I totally believe I'm missing some cases, but there's not enough data for me to see what that is.

Can you send me an example data set that behaves strangely? 

My results:

With the harmonic-oscillator.py example:

Example: pymbar4 without bootstrap:
Uncertainty in free energies is:
[[0.         0.07100713 0.08109591 0.08450777 0.08621464 0.08859106]
 [0.07100713 0.         0.03654687 0.04359901 0.04684609 0.05106776]
 [0.08109591 0.03654687 0.         0.01988063 0.02659403 0.03325712]
 [0.08450777 0.04359901 0.01988063 0.         0.0150578  0.02393041]
 [0.08621464 0.04684609 0.02659403 0.0150578  0.         0.01018445]
 [0.08859106 0.05106776 0.03325712 0.02393041 0.01018445 0.        ]]

With 50 bootstraps:
[[0.         0.06540844 0.07949936 0.08391709 0.08518872 0.08813963]
 [0.06540844 0.         0.03581663 0.04372777 0.04752009 0.05263798]
 [0.07949936 0.03581663 0.         0.01809706 0.02502033 0.03297847]
 [0.08391709 0.04372777 0.01809706 0.         0.01608443 0.0260132 ]
 [0.08518872 0.04752009 0.02502033 0.01608443 0.         0.01144651]
 [0.08813963 0.05263798 0.03297847 0.0260132  0.01144651 0.        ]]

With the pymbar3 version of the examples, for some reason it's not reproducible from run to run (with or without bootstrapping), so the free energies are not the same, but for the uncertainties.

Without bootstrapping:
[[0.         0.06862103 0.07899439 0.08254269 0.08426874 0.08665758]
 [0.06862103 0.         0.0365022  0.04365517 0.04686046 0.05101214]
 [0.07899439 0.0365022  0.         0.0200684  0.02666277 0.0332094 ]
 [0.08254269 0.04365517 0.0200684  0.         0.01492058 0.02370496]
 [0.08426874 0.04686046 0.02666277 0.01492058 0.         0.01011969]
 [0.08665758 0.05101214 0.0332094  0.02370496 0.01011969 0.        ]]

With 50 bootstraps.
[[0.         0.06540844 0.07949936 0.08391709 0.08518872 0.08813963]
 [0.06540844 0.         0.03581663 0.04372777 0.04752009 0.05263798]
 [0.07949936 0.03581663 0.         0.01809706 0.02502033 0.03297847]
 [0.08391709 0.04372777 0.01809706 0.         0.01608443 0.0260132 ]
 [0.08518872 0.04752009 0.02502033 0.01608443 0.         0.01144651]
 [0.08813963 0.05263798 0.03297847 0.0260132  0.01144651 0.        ]]

Which seems pretty much the same.

mikemhenry:
@zhang-ivy Okay! Updated the PR now to point to `pymbar3`

zhang-ivy:
> I would turn off decorrelation for test cases to make sure there's no difference between versions in terms of the decorrelation (or directly verify that the same number of samples is going into MBAR in both cases).

@mrshirts : I generated `u_kn` and `N_k` once and ran pymbar3 vs pymbar4 on the same samples and am no longer seeing such a large discrepancy in the free energies reported by the the two different branches.

When I examine `u_kn` and `N_k` computed from pymbar3 vs pymbar4:
- generated using openmmtools + pymbar3
u_kn shape: (36, 78084)
N_k: [2169 2169 2169 2169 2169 2169 2169 2169 2169 2169 2169 2169 2169 2169
 2169 2169 2169 2169 2169 2169 2169 2169 2169 2169 2169 2169 2169 2169
 2169 2169 2169 2169 2169 2169 2169 2169]
- generated using openmmtools + pymbar4
u_kn shape: (36, 36)
N_k: [1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1]

This suggests that there are differences in how the decorrelated samples are selected when using pymbar3 vs pymbar4. I investigated this and it looks like the problem arises from openmmtools. [These lines](https://github.com/choderalab/openmmtools/blob/47f59f71aefe574b477a1a4949bd2ef653148c9e/openmmtools/multistate/utils.py#L175-L178) make it so that if the `statisticalInefficiency()` call fails (because it does not exist and instead should be calling `statistical_inefficiency()`), it fails silently. This is something we'll want to keep in mind when we re-factor openmmtools to use pymbar4. 

Sorry for the false alarm! I think this branch is working as expected in comparison to pymbar4's bootstrapping branch.

And just following up regarding a question from one of my previous comments: Since it takes awhile to run the bootstrapping, is there a different solver protocol you'd recommend that's less robust but faster than the default?

mikemhenry:
@zhang-ivy when you get a chance can you look into the failing tests?

mrshirts:
Looks like it has something to do with the return values - the tests are expecting a dictionary to be returned, but it's returning tuples (which is what the original API was).  Maybe some tests were not backported to the old API?

mrshirts:
> And just following up regarding a question from one of my previous comments: Since it takes awhile to run the bootstrapping, is there a different solver protocol you'd recommend that's less robust but faster than the default?

I'm working on this with pymbar4.  Since USUALLY the bootstrap free energies are close to the initial free energies, so you definitelt want to start each of these with the "true" value, rather than all 0 values (which is what the code does).  In pymbar4, I have removed the need to do a couple of steps of self-consistent iteration, and so it can jump straight to Newton-Raphson iterations.  But that is not supported in pymbar3 currently. 

One thing that (I think) still works in pymbar3: if your free energies are linearly overlapping, i.e. you expect that states 0 and 1, 1 and 2, etc. overlap, then you can initialize with "BAR", which gets very close to MBAR very fast. But that shouldn't be hard-coded, since if neighboring states are not neighboring, then it will not work well.

pymbar/tests/test_mbar.py
```diff
@@ -87,6 +87,19 @@ def test_mbar_free_energies():
...
+        results = mbar.getFreeEnergyDifferences(uncertainty_method='bootstrap', return_dict=True)
+        fe = results['Delta_f']
+        fe_sigma = results['dDelta_f']
+        fe, fe_sigma = fe[0,1:], fe_sigma[0,1:]
+
+        fe0 = test.analytical_free_energies()
+        fe0 = fe0[1:] - fe0[0]
+
+        z = (fe - fe0) / fe_sigma
+        eq(z / z_scale_factor, np.zeros(len(z)), decimal=0)
```
zhang-ivy, line 101: @mrshirts : I'm reviewing the test more closely now, and I'm wondering whether this is actually a useful test -- its mostly testing that the mbar solved free energies are close to the analytical free energies. However, the bootstrapped free energy matrix `self.f_k_boots` is never actually used to the free energy, its only used to update the uncertainty. Is there a simple way I add a test that checks the bootstrapped uncertainty is reasonable?

mrshirts: Ah, yeah, these are a port of the old test.  I am still trying to figure out the best way to test the bootstraps for #448 - if you see that PR, since bootstrap error is different in that it needs to be initialized differently.

mrshirts: Oh, I see these are already initialized - the newer pytests are much more extensive. and do things differently than the older tests.

It's doing SOMETHING, in that you calculate the number of standard deviations away the error is. But I'm not exactly what the last test is doing.  It's not a very good test - if it's a good error estimate, Z would be Gaussian normal distributed with standard deviation one, but that's not really testing that.   

My instinct is to focus on making the pymbar4 tests better, though.

zhang-ivy:
Also, just wanted to note that I want to play around with this branch a little bit more before merging.  

I'm interested in generating free energy vs time plots (with bootstrapped uncertainties), which requires computing free energies/uncertainties many times on the same trajectory. This will be expensive, so John recommended I play around with the number of `max_iterations` to see if I can find smaller `max_iterations` that gets us a reasonable uncertainty for t < last iteration and then use the default `max_iterations` when I compute the final free energy estimate (where t = last iteration).

In addition to John's suggestion, I'll also play around with @mrshirts's suggestions: 
- Initialize with a free energy
    - where initial free energy is the non-bootstrapped estimate
    - where initial free energy is BAR estimate
- Skip self consistent iterations
- Initialize with "BAR", which gets very close to MBAR

I just want to make sure that I can speed up bootstrapping sufficiently with the current branch as is, and if not, hopefully would just need to push some minor changes that would allow the speed up.

mrshirts:
> This will be expensive, so John recommended I play around with the number of max_iterations to see if I can find smaller max_iterations that gets us a reasonable uncertainty for t < last

So you just want an approximate free energy?  I wonder if the tolerance of the solution would be better to tweak, as it is a more direct way to get an approximate result. This may be particularly true for the bootstraps, because you want to have a correct free energy to a certain precision (or it would satisfy the equations), but you don't necessarily need to get the distribution as accurately as the free energy.

> where initial free energy is the non-bootstrapped estimate

You definitely want to do this - which I think the #302 PR does already. 

> where initial free energy is BAR estimate.  

This can be done by just using initialize = 'BAR' already, just have to look at the timing. 

>Skip self consistent iterations

You can look at PR #448 to see how I implemented this.

pymbar/mbar.py
```diff
@@ -302,7 +303,33 @@ def __init__(self, u_kn, N_k, maximum_iterations=10000, relative_tolerance=1.0e-
...
+            for b in range(self.nbootstraps):
+                for k in range(K): # randomize within the indices with the same K.
+                    # which of the indices are equal to K
+                    k_indices = np.where(self.x_kindices == k)[0]
+                    # pick new random ones of these K.
+                    new_kindices = k_indices[np.random.randint(int(N_k[k]), size=int(N_k[k]))]
+                    rinit[k_indices] = new_kindices
+
+                self.f_k_boots[b,:] = mbar_solvers.solve_mbar_for_all_states(self.u_kn[:,rinit], self.N_k, f_k_init, solver_protocol)
```
zhang-ivy, line 328: @mrshirts  You mentioned that this PR already initializes this PR with the non-bootstrapped free energy estimate when solving the mbar equation for each bootstrapped free energy matrix, but I don't think it does. `f_k_init` here is a copy of self.f_k before the mbar equation is solved, meaning `f_k_init` is just zeros (see line 309). I can push a commit that moves line 309 to after 310, if you agree that that's the right fix here

mrshirts: Hah, you are right, I was remembering what I did for pymbar4.  It is initializing from the previous bootstrap, which is still decent, but probably slightly worse than starting from the "real" value.   I will upload something in the next few min.

zhang-ivy:
@mrshirts I investigated the aforementioned potential speed ups for pymbar3 bootstrapping with my system of interest and here are my findings: 

I'm using `u_kn` with shape (36, 180000) and timed how long it takes to run 10 bootstrap iterations (just because 1000 would take too long, but i'm hoping that these relative timings hold true even when i use > 10 bootstrap iterations).

No bootstrapping: 28 sec
Default pymbar3 bootstrapping: 184 sec
With 1e-6 tolerance in solve_mbar_once: 120 sec
Initializing with BAR estimator once at the beginning: 149.37 sec
Initializing with BAR estimator for each bootstrap: 335 sec
Skip self consistent iterations: 158 sec

Note that all of these experiments are independent (not aggregated). 

These results mostly match up with your results [here](https://github.com/choderalab/pymbar/pull/448#issuecomment-1183457421) with the exception of your observation that solving MBAR for each bootstrapped matrix is faster when initializing with a BAR estimate -- i'm finding that its not actually worth it to compute the BAR estimate for each bootstrapped matrix. 

So I think the additional features that would be helpful to include in this PR are:
1) Adding an API point to `MBAR()` that allows control of `tol` in `solve_mbar_once` 
2) Skipping self consistent iterations

I'm happy to make these changes if you agree with them!

mrshirts:
Sure, go ahead with adding those.  These two features are both implemented in pymbar4, so you can use that for reference (though they may be getting different enough that you can't just copy and paste)

zhang-ivy:
@mrshirts : Just pushed the 2 features above. Please review when you get a chance!

A couple of questions came up as I was making the changes:
- In pymbar4, why did you remove the sqrt here: https://github.com/choderalab/pymbar/commit/786f9963d0827aea6a074f9057590d051f4d9ab7#diff-f5541a3c868c7773917580b717127325539785ff68b367eebe1427e822c878e6L315 and here https://github.com/choderalab/pymbar/commit/786f9963d0827aea6a074f9057590d051f4d9ab7#diff-f5541a3c868c7773917580b717127325539785ff68b367eebe1427e822c878e6L319
- Don't we want this to be 2 by default? https://github.com/choderalab/pymbar/commit/31695da9a4b98b1726431841e181d59d8a9d5353#diff-f5541a3c868c7773917580b717127325539785ff68b367eebe1427e822c878e6R20

mrshirts:
> In pymbar4, why did you remove the sqrt here

Over-optimization, really.  Comparing which is bigger doesn't require a square root, so it saves a couple of operations (only need to print out the square root for reporting). 

> Don't we want this to be 2 by default? 

I'm not sure which 2 your are talking about in the link this is to.

zhang-ivy:
> I'm not sure which 2 your are talking about in the link this is to.

Sorry, let me clarify. I was referring to `min_sc_iter=0` -- shouldn't this be defaulted to 2 here? 
```python
DEFAULT_SOLVER_PROTOCOL = (
    dict(method="hybr", continuation=True),
    dict(method="adaptive", options=dict(min_sc_iter=0)),
)
```

mrshirts:
> Sorry, let me clarify. I was referring to min_sc_iter=0 -- shouldn't this be defaulted to 2?

Ah, there we solve with hybr first, and then, because continuation is on, it continues with the same values and runs a few iterations of newton-raphson to fully converge.  This is basically since the scipy tolerances are behaving differently from solver to solver.

pymbar/tests/test_mbar.py
zhang-ivy: @mrshirts : Final thing (I think) for this PR -- what do you think of this test to check that the bootstrapping is working properly:

```python
from pymbar import MBAR
from pymbar.testsystems import HarmonicOscillatorsTestCase

# Generate harmonic oscillator with plenty of samples for each state
testcase = HarmonicOscillatorsTestCase()
[x_kn, u_kn, N_k, s_n] = testcase.sample(N_k=[10000, 10000, 10000, 10000, 10000])

# Compute non-bootstrapped uncertainty
mbar = MBAR(u_kn, N_k)
results = mbar.getFreeEnergyDifferences(compute_uncertainty=True, return_dict=True)
stderr = results['dDelta_f'][0, -1]

# Compute bootstrapped uncertainty
mbar_boots = MBAR(u_kn, N_k, nbootstraps=200, solver_tolerance=1e-6, initialize='BAR')
results_boots = mbar_boots.getFreeEnergyDifferences(compute_uncertainty=True, uncertainty_method='bootstrap', return_dict=True)
stderr_boots = results_boots['dDelta_f'][0, -1]

assert stderr_boots < 6 * stderr, f"Boostrapped standard error ({stderr_boots}) is more than six times the unbootstrapped standard error ({6* stderr})"

```

mrshirts:
@zhang-ivy , let me know if anything else is needed here from me!

mrshirts:
Will review a final time later today.  

The plan would be to merge this in, then point master to pymbar4, rather than a pymbar3 version.

I guess one question whether pymbar3 is the correct target to merge to? What are the specific changes here?

mikemhenry:
Right now there isn't a difference between pymbar3 and master
https://github.com/choderalab/pymbar/compare/master...pymbar3

So the idea is that `pymbar3` will be the non-default branch, but a long term support branch that we will maintain.

zhang-ivy:
>I guess one question whether pymbar3 is the correct target to merge to? What are the specific changes here?

And yes, pymbar3 is the correct target -- this branch just allows pymbar3 to handle bootstrapping (with features: 1) adding `solver_tolerance` to the `MBAR()` to allow the user to control the solver tolerance and 2) setting `min_sc_iter` to 0 for bootstrapping)

mrshirts:
OK, looked over it and noticed one improvement - it was not returning the uncertainties when bootstrap was being used and return_dict was false (which was the standard way to access) it.  I uploaded a change.  Assuming it passes tests, I think it's set to go. I can't actually approve the review (since I started the PR), so I'll just go ahead and merge once it passes the tests.

zhang-ivy:
@mrshirts : Great catch! 

Actually, can you hold off from merging it for now? I've been testing this branch on more and more cases and I'm seeing this error occasionally:
```python
WARNING: Did not converge to within specified tolerance.
max_delta = 3.693526e-03, tol = 1.000000e-06, maximum_iterations = 10000, iterations completed = 9999
```
I'm not sure whether its coming from [computing the initial free energy estimate](https://github.com/choderalab/pymbar/blob/bootstrap/pymbar/mbar.py#L312) or from computing the free energy for one of the bootstrapped matrices, so I'll need to investigate more. 
I was also thinking of playing around with the following:
- Try not initializing with BAR
- Try not skipping self consistent iterations for bootstrapped iterations
to see if the error goes away.

mrshirts:
Fundamentally, some data sets are going to be hard.  Note that you can STILL adjust the protocol (at least the original protocol, not the bootstrap one) with the old code, without hard coding anything (see https://github.com/choderalab/pymbar/issues/419#issuecomment-803714103).  And turning off BAR initialization for the bootstraps might could work as well.  If you turn on verbose, you should be able to see if BAR is starting things off better or worse than just the previous free energy.

mrshirts:
Let me know your timing - the goal is to make the switch so that 4.0 is released by tomorrow morning.  @mikemhenry, what else would need to be done at this point to make the master point to pymbar4 and tag it with a version number?  @jchodera wanted to do the LTM release after the release.  

On my side, it's just merging JAX and documentation (plus cleaning up documentation a bit).

mikemhenry:
What I would do @mrshirts is merge pymbar4 into master, then tag + make a GitHub release. I've already created the pymbar3 branch so that will take care of the LTM branch.

mrshirts:
OK, so tonight I will plan on merging other branches into pymbar4, and open a PR to master.  Then we can address issues in the morning with merging that PR?

mikemhenry:
I renamed this branch, this will update your local copy:
```
git branch -m pymbar3 pymbar-3-lts
git fetch origin
git branch -u origin/pymbar-3-lts pymbar-3-lts
git remote set-head origin -a
```

mikemhenry:
@zhang-ivy I know you were on vacation :beach_umbrella:  last week :sunglasses: but are you still testing this PR?

zhang-ivy:
@mikemhenry : I found two things I'd like to address before merging: 
1) Divide by zero warning
```
/home/zhangi/miniconda3/envs/perses-paper/lib/python3.9/site-packages/pymbar-3.0.5-py3.9.egg/pymbar/mbar.py:777: RuntimeWarning: divide by zero encountered in log
  Log_W_nk[:, sa] = np.log(A_n[i, :]) + Log_W_nk[:, l]
```
2) Sometimes it fails to converge: https://github.com/choderalab/pymbar/pull/302#issuecomment-1189503087

Going to try to resolve these by tomorrow -- will keep you updated!

zhang-ivy:
Update on the above 2 issues: 

Re the divide by 0 warning, @mrshirts [helped me identify the problem](https://github.com/choderalab/pymbar/pull/448#issuecomment-1196845293). I will open a PR to fix this (and we can merge it into pymbar-3-lts after we merge `bootstrap` in)

Re the occasional convergence failures, I found that when I do not feed `initialize='BAR` to `MBAR()`, the problem goes away. Moving forward, I won't initialize my free energy with the BAR estimate.

zhang-ivy:
@mikemhenry : I think this branch is ready to be merged!

PR #308, opened 2019-01-30
Fixed README.md for examples

mrshirts:
Fixed README so the examples use the correct calling instructions.


mrshirts:
Any thoughts on the errors here?  They appear to be unrelated to the changes (which are all just changes to the README.md.

Lnaden:
Yes. It looks like the pinning to the old Conda Build version from the pre-build shell file is throwing an error since it won't work on a newer version of base Conda. I think simply removing the pin from the conda-build version should work since this package is not deployed to omnia anymore, and instead just conda-forge

mrshirts:
I'm going to merge, since it's just text, and it's independent of the conda issue.

PR #309, opened 2019-03-26
[WIP] Separate pmf

mrshirts:
Work in progress for discussion. First pass at separating out the PMF code from MBAR.  

Creates a class that separates out the creation of the PMF from the evaluation of the PMF at a range of coordinates. 

Reproduces (up to formatting errors and convention choices) the 4 examples in the repository in 1D and 2D. 

Will make it much easier to add PMF methods to the code in the future.

PR #310, opened 2019-03-27
[WIP] adding a kernel density estimate to PMF

mrshirts:
[WIP], based on #309. 

With the change to a module, it becomes much easier to add new PMF methods. As an example, I have added kernel density estimation.  We take all of the biased samples, with their weights, and construct a kernel density estimate, using sklearn.   

Implemented in harmonic_oscillators.py; now both standard histogram and sklearn-based kde are implemented there. 

Does not yet have uncertainty in the KDE estimator; I figure it will be better to do with bootstrap.


mrshirts:
Replaced by #327

PR #312, opened 2019-05-11
Changed heat capacity example ('u_kln' to 'u_kn')

garrettameek:
Restructured the heat capacity example to build 'u_kn' instead of 'u_kln' input arrays for pymbar.


mrshirts:
Switch back to dictionaries for the results; far more pythonic that using arrays.  Much easier to ask for the the means and standard deviations than index 0 and 1 - one has to remember what those mean?  Dictionaries are self-documenting.

PR #313, opened 2019-05-15
Changed heat capacity example ('u_kln' to 'u_kn')

garrettameek:
Restructured the heat capacity example to build 'u_kn' instead of 'u_kln' input arrays for pymbar.  In contrast to my last pull request, this update maintains dictionary-style indexing of the pymbar output arrays.


mrshirts:
#394 used instead of this.

PR #314, opened 2019-11-11
Python 3.8 Conda-Forge Fixes

Lnaden:
In order to accomodate conda-forge/pymbar-feedstock#5 this PR fixes the travis-ci tests (in theory) and then we can do a 3.0.4 release which has the fixes in #296 which the C-F build needs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,8 +32,8 @@ script:
 env:
   matrix:
     - python=2.7  CONDA_PY=27
-    - python=3.5  CONDA_PY=35
-    - python=3.6  CONDA_PY=36
+    - python=3.7  CONDA_PY=37
+    - python=3.8  CONDA_PY=38
 
   global:
     - ORGNAME="conda-forge"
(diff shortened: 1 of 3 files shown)


Lnaden:
Keep this open for now as we are waiting for https://github.com/conda-forge/statsmodels-feedstock/pull/17 to merge. I'm watching it, once its done and through the c-f cdn, we can re-run the Python 3.8 builds. Otherwise its passing.

jchodera:
@jaimergp highly recommends we look at GitHub Actions, since they offer free Azure CI for linux, win, and osx using simple pre-built recipes from their marketplace:
https://github.com/features/actions

It's supposedly also super-fast, and would let us ditch paying for AppVeyor.

Might be worth considering for the MolSSI Cookiecutter.

Lnaden:
>  recommends we look at GitHub Actions

I've been looking into these before today and it might be something we switch to, the integrate well with Azure since Microsoft owns both of them now. 

For now though, I think just merging this and making a patch release so we can cut the conda-forge release is what we need right now. We can iterate on the rest later.

PR #316, opened 2019-11-19
BAR returns a dict now

panxl:


jchodera:
Thanks for contributing this!

What would you think about also adding a test case that would have caught the problem with BAR-based initialization?

panxl:
Thank you for the suggestion!

I added a test case to initialize MBAR with BAR initializer, but I am not sure if it is appropriate or not since I am not very familiar with Python testing.

PR #319, opened 2019-12-01
Make all outputs default to 3.0.3 behavior

### LINKED ISSUE #318: 3.0.4 bugfix release accidentally contains API-breaking horrors
cc: https://github.com/choderalab/openmmtools/issues/452#issuecomment-560135185
cc: @jaimergp @Lnaden @mrshirts @andrrizzi 

It looks like the 3.0.4 bugfix release accidentally contains API-breaking horrors which breaks everything that uses it:
* Looks like `3.0.4` [uses a `dict` to store the results](https://github.com/choderalab/pymbar/compare/3.0.3...3.0.4#diff-2f68bad712f6798ec69aced2640b2215R218) and returns that. 
* `3.0.3` returned [either a tuple or a single value](https://github.com/choderalab/pymbar/compare/3.0.3...3.0.4#diff-2f68bad712f6798ec69aced2640b2215L376).

We need to revert this and save API-breaking changes for a 4.0.0 release so people have time to change their code.

Here's my suggested plan:
* Every code that uses `pymbar` will break if we leave 3.0.4 in the wild, so I suggest we remove it from `omnia` and `conda-forge` channels. I can wipe the `omnia` package, but we can't easily wipe the [conda-forge package](https://anaconda.org/conda-forge/pymbar/files), so we will have to ask them to move this to a broken package label. 
* It seems like way more got swept up in 3.0.4 than intended, so we should either re-release with cherry-picked commits for python 3.8 and scipy 1.0 compatibility, or issue a 3.0.5 that just has those changes but does not contain newer changes
* Anything API-breaking should be introduced as an optional mode, where an optional kwarg could specify the new API. That allows folks a path to try the new API while not breaking old code. But I suggest we save that for 3.1.0.
* Any permanent API-breaking changes must have a major version increment: Moving to a `dict` as standard would be in pymbar 4.

We'll move ahead with this plan on Monday unless there are objections, since this has broken everything in our free energy ecosystem.

Huge thanks to @jaimergp for noticing what happened!

Lnaden:
Still allows dictionary output via kwarg and tests have been added to
ensure the two outputs are the same. This should be considered a
candidate for 3.0.5 release.

Fixes #318
Fixes #316

cc: https://github.com/choderalab/openmmtools/issues/452#issuecomment-560135185
cc: @jaimergp @Lnaden @mrshirts @andrrizzi

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,14 +45,15 @@ To analyze this data, we first initialize the `MBAR` object:
 ```
 Estimating dimensionless free energy differences between the sampled thermodynamic states and their associated uncertainties (standard errors) simply requires a call to `getFreeEnergyDifferences()`:
 ```python
->>>  results = mbar.getFreeEnergyDifferences()
+>>>  results = mbar.getFreeEnergyDifferences(return_dict=True)
 ```
 Here `results` is a dictionary with keys `Deltaf_ij`, `dDeltaf`, and `Theta`. `Deltaf_ij[i,j]` is the matrix of dimensionless free energy differences `f_j - f_i`, `dDeltaf_ij[i,j]` is the matrix of standard errors in this matrices estimate, and `Theta` is a covariance matrix that can be used to propagate error into quantities derived from the free energies.
+By default, `return_dict` is `False` and the return will be a tuple.
 
 Expectations and associated uncertainties can easily be estimated for observables `A(x)` for all states:
 ```python
 >>> A_kn = x_kn # use position of harmonic oscillator as observable
->>> results = mbar.computeExpectations(A_kn)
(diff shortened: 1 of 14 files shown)


jchodera:
Thanks so much, @Lnaden! We can go with a 3.0.5 release to fix this if we roll it out rapidly.

Lnaden:
If this is everything you wanted for the 3.0.5, it should be ready to go. I don't think the Windows failures are because of the repo, and more because of conda and DLL linking in some of the recent versions. Also because Python 2 which we are dropping soonish anyways.

jchodera:
New release: https://github.com/choderalab/pymbar/releases/tag/3.0.5

PR #324, opened 2020-01-28
Fixes sanity check for empty bins in computePMF

tuckerburgin:
Previous commit uses "if (len(indices) == 0)" to check that the bin is not empty; however, "indices" is a one-length tuple of which the first (and only) element is the list of indices in that bin, causing this check to always pass even for empty bins (the length of indices can only ever be one). Adding "[0]" causes the test to behave as intended.


Lnaden:
Good catch.

PR #325, opened 2020-02-02
Fixed return type of multiple values to dict-like object

mrshirts:
First pass at deprecating tuple return types. To keep things roughly backwards compatible, I am proposing that we leave in return_dict, but change the default to False.  Thus, if people want previous behavior, they just need to add a flag, rather than rewriting the code.  

With return_dict=False, however, it will now give a user error that the feature will be removed in future releases.  I used UserWarning rather than DeprecationWarning, since the latter are only printed in developer types, and we want the users to know they should change the way they call it.

Currently, the return is STILL a dict, and the variable is return_dict. I could change return_dict to return_object or something like that, since we may change it into a dict-like object in the future, but I don't know if that is needed now.  We may just drop return_dict before we get around to replacing the dict with a dict-like object.

Still need to change the examples to remove the return_dict keyword, since they already use the new default.

### DIFF
--- a/pymbar/exp.py
+++ b/pymbar/exp.py
@@ -51,8 +51,8 @@ from pymbar.utils import logsumexp
 # One-sided exponential averaging (EXP).
 #=============================================================================================
 
-def EXP(w_F, compute_uncertainty=True, is_timeseries=False, return_dict=True):
-    """Estimate free energy difference using one-sided (unidirectional) exponential averaging (EXP).
+def EXP(w_F, compute_uncertainty=True, is_timeseries=False):
+    """Estimate free energy difference using one-sided (unsidirectional) exponential averaging (EXP).
 
     Parameters
     ----------
@@ -63,17 +63,13 @@ def EXP(w_F, compute_uncertainty=True, is_timeseries=False, return_dict=True):
     is_timeseries : bool, default=False
         if True, correlation in data is corrected for by estimation of statisitcal inefficiency (default: False)
         Use this option if you are providing correlated timeseries data and have not subsampled the data to produce uncorrelated samples.
-    return_dict : bool, default True
-        If true, returns are a dictionary, else they are a tuple
 
     Returns
     -------
     'Delta_f' : float
         Free energy difference
-        If return_dict, key is 'Delta_f'
     'dDelta_f': float
         Estimated standard deviation of free energy difference
-        If return_dict, key is 'dDelta_f'
 
     Notes
     -----
@@ -96,7 +92,6 @@ def EXP(w_F, compute_uncertainty=True, is_timeseries=False, return_dict=True):
     """
 
     result_vals = dict()
-    result_list = []
 
     # Get number of work measurements.
     T = float(np.size(w_F))  # number of work measurements
@@ -128,17 +123,10 @@ def EXP(w_F, compute_uncertainty=True, is_timeseries=False, return_dict=True):
         # Return estimate of free energy difference and uncertainty.
         result_vals['Delta_f'] = DeltaF
         result_vals['dDelta_f'] = dDeltaF
-        result_list.append(DeltaF)
-        result_list.append(dDeltaF)
     else:
         result_vals['Delta_f'] = DeltaF
-        result_list.append(DeltaF)
-    if return_dict:
-        return result_vals
-    else:
-        warnings.warn("Returning values as a tuple is deprecated, and will be removed in a future version.",
-                      UserWarning)
-        return tuple(result_list)
+
+    return result_vals
 
 #=============================================================================================
 # Gaussian approximation to exponential averaging (Gauss).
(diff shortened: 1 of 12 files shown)


mrshirts:
Whoops, forgot that other files have return_dict as well as mbar.py.  Will add those now.

Lnaden:
I need to double check all the uses, but this is on the right track! Do we want to warn people about the returns now, or just make the change since we are bumping the major version anyways?

mrshirts:
> Do we want to warn people about the returns now, or just make the change since we are bumping the major version anyways?

I guess it's best practices to warn before removing functionality?  Now, it breaks but they know what they need to do to fix it, rather than failing silently.

Lnaden:
> I guess it's best practices to warn before removing functionality?

Yes it is. But since we're doing the major bump, we may just want to commit to it now. What if we put the warning in *now* and then cut a minor revision in a 3.0.X+1? I don't think 4 is going to be done for a few months anyways, so it should have enough time to give people warning.

jchodera:
I'm with @lnaden: Let's just fix everything at this major version update, including fixing the API for PEP8. Let's aim for maintainable and a clean, usable API, rather than backwards compatibility.

Do we want to call this something other than `pymbar`?

mrshirts:
> Do we want to call this something other than pymbar?
Eh. no.  We have a lot of name recognition.  Just bump the version number, we can emphasize the 4 in release materials/documentation.

Lnaden:
> Do we want to call this something other than pymbar?

I'm with Michael on this one, keep the PyMBAR name.

mrshirts:
> Yes it is. But since we're doing the major bump, we may just want to commit to it now. What if we put the warning in now and then cut a minor revision in a 3.0.X+1? I don't think 4 is going to be done for a few months anyways, so it should have enough time to give people warning.

I think that the people whose code it's most likely to break will be people that wrote scripts a year or so ago, so a minor fix won't help much.

I'm OK with a full change.  We'll just need to emphasize it heavily in the release notes.

mrshirts:
OK, did a full stripping, including examples and tests.

pymbar/exp.py
```diff
@@ -51,8 +51,8 @@
 # One-sided exponential averaging (EXP).
 #=============================================================================================
 
-def EXP(w_F, compute_uncertainty=True, is_timeseries=False, return_dict=False):
-    """Estimate free energy difference using one-sided (unidirectional) exponential averaging (EXP).
+def EXP(w_F, compute_uncertainty=True, is_timeseries=False):
+    """Estimate free energy difference using one-sided (unsidirectional) exponential averaging (EXP).
```
jaimergp: Is this a typo?

pymbar/mbar.py
```diff
@@ -404,27 +405,25 @@ def computeEffectiveSampleNumber(self, verbose = False):
...
-    def computeOverlap(self, return_dict=True):
+    def computeOverlap(self):
         """
         Compute estimate of overlap matrix between the states.
 
         Parameters
         ----------
-        return_dict : bool, Default False
-            If true, results are a dict, else a tuple
+        None
```
jaimergp: If there are not parameters, I think we can remove this section from the docstringl

pymbar/mbar.py
```diff
@@ -404,27 +405,25 @@ def computeEffectiveSampleNumber(self, verbose = False):
...
 
         Returns
         -------
+        Results dictionary with the following keys:
```
jaimergp: Double check if this syntax gets properly recognized in the rendered documentation.

pymbar/mbar.py
```diff
@@ -482,22 +479,19 @@ def getFreeEnergyDifferences(self, compute_uncertainty=True, uncertainty_method=
             than this number (default: 1.0e-10)
         return_theta : bool, optional
             Whether or not to return the theta matrix.  Can be useful for complicated differences.
-        return_dict: bool, default False
-            If true, returns are in a dictionary otherwise a tuple is returned
 
         Returns
         -------
+        Results dictionary with the folowing keys:
```
jaimergp:
```suggestion
        Results dictionary with the following keys:
```

pymbar/exp.py
mrshirts: Yep, not sure how got there.

pymbar/mbar.py
mrshirts: Eh, I think it helps to be explicit?

pymbar/mbar.py
mrshirts: I'm not actually sure how to check the rendered documentation.  What should I run to get it?

pymbar/mbar.py
mrshirts: Done.

mrshirts:
There's the question if the documentation displays correctly, but probably the documentation hasn't really been updated in 3 years, and I'm not sure even builds right now, which is probably something that should be done separately.

mrshirts:
I'll go ahead and merge for now, then.

PR #327, opened 2020-02-09
Revamping of PMF

mrshirts:
Here's my proposed revamping of PMF functionality.  I'd love to get comments so we can merge this in for pymbar4.  I anticipate there may be some requested changes, so I'd love to get started on them.


mrshirts:
> Is the need for the split of the PMF capabilities because you wanted them to be expanded? 

There is  a lot there, and a lot more stuff that could go in the future.  It's certainly more different from mbar.py as exp.py and bar.py!  On the basis of modularization, it makes sense to have all the related methods in a different file, since there is little functional overlap in the techniques.

Be more specific what I should look at when you say "check the test coverage".

Lnaden:
> Be more specific what I should look at when you say "check the test coverage".

I mean we should make sure the tests are updated in a way to maximally hit lines of code, especially since this is a new module. The existing tests I think just run through the PMF computation, but not necessarily  all the functions and code paths. 

We can do this in another PR to avoid making this one too large, especially since I want to change from Nose -> Pytest

mrshirts:
> We can do this in another PR to avoid making this one too large, especially since I want to change from Nose -> Pytest

OK, added issues to switch  testing frameworks (#330, @Lnaden ), and after to add more PMF tests (#331 @mrshirts )

Lnaden:
I'm of the opinion lets get this in, then we can work on creating tests and whatnot after I handle #330. I have some time today so I'd like to get the test framework switched over, and start on #332 as well.

Since this is all on the `pymbar4` branch, I feel safe making rapid changes now, improve the CI/test/coverage framework, then going back and ensuring we're covering all our bases.

mrshirts:
@jchodera Do you want to take an overall look at how this is structured? There's a couple of things that need to be done with the usability of the MCMC sampling of parameters, but it would be good to know if there are any showstoppers about the general usability and way it is presented.  Check the examples to see how PMF's are now generated for most users.

mrshirts:
mbar_computePMF test obviously failing.    

Should be completely fixed in the near future by adding a test_pmf.py file (test_bar.py and test_exp.py need to be done as well!), but for now I have removed the broken function call.

Lnaden:
Great! If this is good to go in your book, go ahead and merge!

mrshirts:
Note: address issue with a deprecated KDE functionality being used.

mrshirts:
OK, failed because the conda-recipe (I think?) doesn't have sklearn.  So it will need to be added to run kernel density calculations.

mrshirts:
Let's go with merging this now with basic tests now implemented?

PR #328, opened 2020-02-11
Move CI to GHA and start upgrade to Cookiecutter style

Lnaden:
As it says, this (assuming it works) ~should also go into pymbar4 branch.~

Edit: if merged, it will go into pymbar4 branch first and should still work.


Lnaden:
So GHA is awesome. That is all. I think this should be ready and once its merged. I changed the base to `pymbar4` as the PR target, once its in, I think I have it set up correctly to trigger anyways. I'm still running the GHA on my local fork. You can see them here: `https://github.com/Lnaden/pymbar/actions/runs/38200430` 

In theory, this should replace the CI for Travis and Appveyor

jaimergp:
LGTM!

Do we want nightlies too? You can add them [like this](https://github.com/choderalab/perses/blob/master/.github/workflows/CI.yaml#L6-L8).

Lnaden:
Sure!

PR #334, opened 2020-02-14
Replace `np.matrix` calls and other warnings

Lnaden:
This PR is an extension of #333, so it is ~DNM until 333 is done.~ (Ready)

I will refresh this once #333 is to fix the edits, so if you want to hold off on review, thats fine too.

This PR replaces all of the `np.matrix` calls and manipulates the
objects as arrays only. See https://numpy.org/doc/1.18/user/numpy-for-matlab-users.html
for why this needed to be done.

However, that does mean there are some syntax changes I had to make since
`*` on arrays is element-wise (matrix multiplication can be done with
`@` operator since Python 3.5, though you can do `np.matmul(a,b)` to
the same effect ).
And transpose of 1-D arrays doesn't do anything like it does to a matrix, but I found
doing `x - np.astleast_2d(x).T` does, while still being arrays and not
matrices. Note: this is MUCH safer than assuming its 1-D and doing `x - x[np.newaxis].T` 
which will cause very silly things to happen if `x` is already 2-D.

Edit: the `vstack` is better than `atleast_2d`, thanks mrshirts for that suggestion.

That said, there's alot of replacements here and I'm hoping someone
else can look over this and make sure I have the correct matrix operations
(namely element-wise vs matrix-style) in place for the correct results.
This does pass my local tests.

### DIFF
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -6,7 +6,7 @@ dependencies:
     # Base depends
   - python
   - pip
-  - numpy >=1.11
+  - numpy >=1.12
   - scipy
   - numexpr
   - six
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -14,7 +14,7 @@ requirements:
     - toolchain
   run:
     - python
-    - numpy >=1.11
+    - numpy >=1.12
     - scipy
     - numexpr
     - six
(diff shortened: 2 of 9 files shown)


mrshirts:
See this PR to Lnaden/pymbar (#1) to clean this matrix code up additionally by removing some unneeded type casting and doing x + x.T with a simpler x + np.vstack(x).

mrshirts:
>  There may be opportunities to use opt_einsum for both performance and simplicity.

I think things should be relatively simple now.  I'm fairly certain that the logsum is by far the limiting factor, so accelerating that is the biggest bottleneck, and should be dealt with before anything else is tried.

Lnaden:
I made a git typo on this, fixing.

Lnaden:
This should be fixed now

mrshirts:
I'd say go ahead and merge.

PR #335, opened 2020-02-16
Shorten duplicate state check

mrshirts:
Reduce the cost of checking for duplicate states. Previously, all of the samples was compared, which is unnecessary; this PR has it just check the first 50 samples in all states (unless there are les than 50 total samples), which is much faster.  This is almost certainly sufficient in essentially all cases.  This is an optional test that only occurs with verbose on.

This could potentially be made shorter and a little faster with clever use of numpy.unique(), but it's not clear that would really help that much besides shortening the number of lines code, since with this change, the speed cost is mostly removed. numpy unique does an unnecessary sort as well.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -1,0 +1,73 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - 'master'
+      - 'pymbar4'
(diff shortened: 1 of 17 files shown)


mrshirts:
Committed to wrong branch.

PR #337, opened 2020-02-16
Reduce the cost of check for duplicate states

mrshirts:
Reduce the cost of checking for duplicate states. Previously, all of the samples was compared, which is unnecessary; this PR has it just check the first 50 samples in all states (unless there are les than 50 total samples), which is much faster. This is almost certainly sufficient in essentially all cases. This is an optional test that only occurs with verbose on.

This could potentially be made shorter and a little faster with clever use of numpy.unique(), but it's not clear that would really help that much besides shortening the number of lines code, since with this change, the speed cost is mostly removed. numpy unique does an unnecessary sort as well.

### DIFF
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -226,16 +226,30 @@ class MBAR:
 
         # perform consistency checks on the data.
 
-        # if, for any set of data, all reduced potential energies are the same,
-        # they are probably the same state.  We check to within
-        # relative_tolerance.
-
         self.samestates = []
         if self.verbose:
+
+            # if, for any set of data, all reduced potential energies are the same,
+            # they are probably the same state.
+            #
+            # This can take quite a while, so we do it on just
+            # the first few data points.
+            #
+            # determine if there are less than 50 points to compare energies.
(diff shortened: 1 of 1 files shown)


mrshirts:
For whatever reason, seem to be failing on a test_convariance test?

Lnaden:
That looks like its happening due to a bug which was still around when we still had list returns, I think you need to do a git rebase of this from the pymbar4 branch and that will fix itself.

mrshirts:
Thanks, tried again!

mrshirts:
OK, works now!

Lnaden:
As I mentioned in the review, what if we drew a random 50 indices instead of the first 50? I think it would a simple thing which adds to the robustness of the check. Would that add too much overhead for what this fixes?

mrshirts:
I'll give it a try later today.

mrshirts:
OK, did a version with random 50!

mrshirts:
Based on previous approval, merging.

Lnaden:
The coverage report should get less noisy once we merge into master. I'll see about making it quieter

> OK, did a version with random 50!

Groovy

PR #339, opened 2020-02-16
test_covariance appeared to be testing the wrong thing.

mrshirts:
I fixed some issues in test_covariance that didn't quite address the return type problem for results.


mrshirts:
Something weird going on, will check.

PR #342, opened 2020-02-17
Fix formatting

mrshirts:
Fix formatting for the code to move to the .format functionality instead of printf-like formatting.

### DIFF
--- a/examples/constant-force-optical-trap/extract-data.py
+++ b/examples/constant-force-optical-trap/extract-data.py
@@ -28,7 +28,7 @@ except ImportError:
 
 original_data_directory = 'original-data' # location of original data (.xls files)
 processed_data_directory = 'processed-data' # location for where processed data is to be stored
-datasets = ['20R55_4T'] # prefixes of datasets to process (filenames are '%(prefix)_data.xls')
+datasets = ['20R55_4T'] # prefixes of datasets to process (filenames are ',format((prefix)_data.xls')
 
 #=============================================================================================
 # SUBROUTINES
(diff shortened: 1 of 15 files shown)


mrshirts:
OK, clearly didn't fix the bad escape characters.

mrshirts:
OK, that got it.  Got all escape characters.  Just need to fix matrix errors which will be fixed by #334

examples/constant-force-optical-trap/extract-data.py
```diff
@@ -40,56 +40,56 @@
...
-    print("Extracting data from dataset '%s'..." % dataset)
+    print("Extracting data from dataset '{}'...".format(dataset))
 
     # Open Excel spreadsheet file.
     workbook_filename = os.path.join(original_data_directory, dataset + '_data.xls')
     workbook = xlrd.open_workbook(workbook_filename)
 
     # DEBUG
-    print("Workbook contains %d worksheets:" % workbook.nsheets)
+    print("Workbook contains {:d} worksheets:",format(workbook.nsheets))
```
jaimergp: This should be `.format(...)` (notice the period instead of the comma)

examples/constant-force-optical-trap/extract-data.py
```diff
@@ -28,7 +28,7 @@
 
 original_data_directory = 'original-data' # location of original data (.xls files)
 processed_data_directory = 'processed-data' # location for where processed data is to be stored
-datasets = ['20R55_4T'] # prefixes of datasets to process (filenames are '%(prefix)_data.xls')
+datasets = ['20R55_4T'] # prefixes of datasets to process (filenames are ',format((prefix)_data.xls')
```
jaimergp: `s/,/./` :)

Lnaden:
We don't support less than 3.6 (3.7 now technically, although we are still 3.6 compatible.), so either `format` or `f-strings`, and it would not be too hard to replace them all.

mrshirts:
So, if both .format and f-strings are accepted for 3.6 (and the medium-future), I'd rather stick with .format for now.  f strings are not necessarily more readable to me!  I just got used to .format instead of printf-like behavior!  

If f-string are now preferred, I could be swayed.

examples/constant-force-optical-trap/extract-data.py
mrshirts: Yeah, I wasn't bothering with ones in comments :)

examples/constant-force-optical-trap/extract-data.py
mrshirts: Thanks, will fix, I think that this code didn't execute because of missing a dependency, and I didn't worry, since the example runs without it (data already extracted).

jaimergp:
`.format()` will be there in the foreseeable future, so no problems with that. 

Although `f-strings` have better performance, it doesn't make much of a difference in practice. I guess it's a matter of personal preference, so it's totally ok if you want to stick to `.format()`.

Lnaden:
The performance should not mater here for `f-string` vs `.format()` since we are not doing that much string formatting in process. So Dealers Choice!

mrshirts:
@Lnaden , @jaimergp , look good now?

PR #347, opened 2020-02-17
Start Coverage Improvements (Mbar.py)

Lnaden:
This PR starts the coverage improvements for the package as a whole.
The focus here is on the mbar.py file as well as some initial ignores
for the CodeCov configs.

I also found and fixed 2 bugs with this:
- `svd` method missed a `@` operator (introduced in #334 by me)
- "BAR" init method failed due to dict return.

One outstanding question I have is the `computeCovarianceOfSums`
function. It's not used anywhere in the code and the last time it was
used was some time ago (I can go find the commit again if need be).
GitHub also does not seem to think anyone other than PyMBAR clones
use the function either.
If we want to keep this function around, is there a decent test which
can be written for it? The one I added simply runs it, not sure if its
accurate though.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,5 +1,11 @@
 # Codecov configuration to make it a bit less noisy
 coverage:
+  ignore:
+    - pymbar/tests/.*
+    - pymbar/testsystems/.*
+    - setup.py
+    - pymbar/version.py
+    - pymbar/old_mbar.py
   status:
     patch: false
     project:
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -72,7 +72,7 @@ class MBAR:
     # =========================================================================
 
     def __init__(self, u_kn, N_k, maximum_iterations=10000, relative_tolerance=1.0e-7, verbose=False, initial_f_k=None,
-                 solver_protocol=None, initialize='zeros', x_kindices=None, **kwargs):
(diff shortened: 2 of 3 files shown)


mrshirts:
> I also found and fixed 2 bugs with this:

Yay!

> One outstanding question I have is the computeCovarianceOfSums

This should probably be used to generate entropy and enthalpy, but it was never added like I intended. I'll adjust it to do that.  It's also useful for a number of other more complex observable applications (like a Delta Delta G).  I can add some examples to the harmonic-oscillator.py.

> the last time it was used was some time ago (I can go find the commit again if need be).

If you can easily find it, that would be great.

>If we want to keep this function around, is there a decent test which
can be written for it?

Hmm.  We could explicitly check externally it does the same things it does internally.  What else are you thinking of?

Lnaden:
This is the last time the function was used (specifically when it was removed):

https://github.com/choderalab/pymbar/commit/a46640827a7c6f0530a64d685cd03602c8adbb20

I don't know exactly when it was merged into `master`

Lnaden:
> Hmm. We could explicitly check externally it does the same things it does internally. What else are you thinking of?

I'm not actually sure to be honest. Im not sure what a good statistical test we can use to make sure its being calculated correctly would be. At minimum, if we want to keep it, its parameter docstrings could be updated.

Lnaden:
I'm merging this, we can figure out the `covarianceOfSums` function in a later PR

PR #349, opened 2020-02-18
Bar.py and exp.py pytests

mrshirts:
These are adapted from test_mbar.py. Probably could be simpler, but this works.

Also includes one bug fix for bar.py caught by the tests (bad method keyword was not caught and silently returned 0.0


mrshirts:
wrong place to do the PR

Issue #345, opened 2020-02-17
Add pytest tests for bar.py and exp.py


mrshirts:

mrshirts:
Addressed in #350

mrshirts:
Fixed with #350

PR #350, opened 2020-02-18
 Bar.py and exp.py tests

mrshirts:
These are adapted from test_mbar.py. Probably could be simpler, but this works.

Also includes one bug fix for bar.py caught by the tests (bad method keyword was not caught and silently returned 0.0).

Addresses #345

### DIFF
--- a/pymbar/bar.py
+++ b/pymbar/bar.py
@@ -224,6 +224,12 @@ def BAR(w_F, w_R, DeltaF=0.0, compute_uncertainty=True, uncertainty_method='BAR'
         method = 'self-consistent-iteration'
         DeltaF_initial = DeltaF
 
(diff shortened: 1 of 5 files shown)


Lnaden:
You can change the base the PR is being proposed into by using the `edit` button at the top here of GitHub, you don't have to close and re-open if you don't want to.

mrshirts:
Looks like failures to check out the code?  Rerunning to see if that fixes it.

mrshirts:
No, seems to be failing on checkout for some of the tests?  @Lnaden, and ideas?

mrshirts:
OK, was some weird mis-sync in the loading.  All good now.

Lnaden:
Yeah, that was a new one for me too. Glad to see it fixed itself.

Issue #351, opened 2020-02-18
Address collections warning in testing?


mrshirts:
Tests are generating following warning. Collections isn't in the code as far as I can tel.  Is the conda package that the testing runs just out of date?  What else would need to be addressed to fix?

/usr/share/miniconda/envs/test/lib/python3.8/site-packages/patsy/constraint.py:13
  /usr/share/miniconda/envs/test/lib/python3.8/site-packages/patsy/constraint.py:13: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
    from collections import Mapping

-- Docs: https://docs.pytest.org/en/latest/warnings.html

----------- coverage: platform linux, python 3.8.1-final-0 -----------
Coverage XML written to file coverage.xml

Lnaden:
That is an upstream package `patsty` which is installed by `statsmodels` throwing that warning. Not much we can do about it directly since its on the `patsy` devs to fix it. I can look into suppressing warnings from other packages we don't use directly since I dont know if/when they will update it.

mrshirts:
I guess mark this as `won't fix` for now?

jaimergp:
`patsy` got this addressed in https://github.com/pydata/patsy/pull/140#issuecomment-433760027, but they haven't cut a release yet. Issue asking for that is here: https://github.com/pydata/patsy/issues/154. They are specifically asking for the release to remove the `collections` warning.  Let's see if we can add some pressure there :)

jaimergp:
We _could_ (I am not saying this is a good idea) manually upgrade `patsy` to the `master` release to remove the warnings in our CI if this is really annoying us.

Lnaden:
This will be fixed (at least suppressed) in #357, we can undo it later.

Lnaden:
Fixed by #357

PR #357, opened 2020-02-20
Even more coverage! [WIP]

Lnaden:
This PR works to bring even more coverage of tests. I got utils coverage all the way up. However, I want to discuss the plan for PMF and some of the other choices I plan on making:

* The PMF module has many code paths which I can explore and test. Do we want to run down all of them? 
* For PMF methods without analytical errors, do we want to rely on the bootstrap to test them? I started doing this, but adding many bootstraps will significantly slow down the process.
* My IDE caught a bunch of syntax typos which were not hit in the tests, I'll try to fix them as I run into them.
* The Spline PMF test makes this call: 
    ```python
    pmf.generatePMF(u_kn, x_n,...)
    ```
    which uses `u_kn` instead of `u_n` like the histogram and KDE tests. I don't know if that was a typo so I wanted to ask.
* I assumed `numexpr` is required since we do require it in both the Conda env/recipe and in the setup.py
* I dropped the old `get_fn` which wasn't used and I think left over from copying the MDTraj utilities a long while back
* I plan on completely dropping `six` as a dependency, any objections?

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -9,3 +9,8 @@ description-file = README.md
 
 [aliases]
 test = pytest
+
+[tool:pytest]
+filterwarnings =
+    ignore::DeprecationWarning:patsy.*
+    ignore::PendingDeprecationWarning
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,8 @@ pip-log.txt
 .coverage
 .tox
 nosetests.xml
+coverage.xml
+htmlcov
 
 # Translations
 *.mo
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -16,4 +16,5 @@ dependencies:
   - pytest-cov
   - codecov
   - statsmodels
+  - scikit-learn  # Provides the sklearn module
 
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -9,7 +9,6 @@ dependencies:
   - numpy >=1.12
   - scipy
   - numexpr
-  - six
 
     # Testing
   - pytest
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -9,7 +9,6 @@ requirements:
   build:
     - python
     - setuptools
-    - six
     - numpy 1.11.*
     - toolchain
   run:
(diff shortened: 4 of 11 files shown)


mrshirts:
> The PMF module has many code paths which I can explore and test. Do we want to run down all of them?

Some are more experimental, so I don't know that one needs to look at ALL of them.  I'll look over this now that you have updated the pytest stuff.

> For PMF methods without analytical errors, do we want to rely on the bootstrap to test them? I started doing this, but adding many bootstraps will significantly slow down the process.

I think that is very expensive.  My kludge is to use the histogram errors (which are fast), and should be reasonable for the amount of data collected. 

>which uses u_kn instead of u_n like the histogram and KDE tests. I don't know if that was a typo so I wanted to ask.

I don't think so, but I will check later when I look over the tests and explain why it's not  an error.

> I assumed numexpr is required since we do require it in both the Conda env/recipe and in the setup.py
It's used for logsumexp which is called by MBAR so I think so?

> I dropped the old get_fn which wasn't used and I think left over from copying the MDTraj utilities a long while back

Sounds good.

> I plan on completely dropping six as a dependency, any objections?

Sounds good.

jaimergp:
> I plan on completely dropping six as a dependency, any objections?

Please 🙏

Lnaden:
I think this is at the point where I've done what I can for the first pass. Theres lots of PMF codepaths which don't work for various reasons, and instead of having the tests fail for now, I just mark them `xfail` and we can circle back later. I've also cut back on the number of test permutations for speed sake, namely the `nbootstraps` which was adding a slow error computation to the KDE and Spline methods. 

This does Fix #351 in that I suppress the CodeCov warning for that collection from Patsy.

Six is no longer part of the repo now, yay!

mrshirts:
> I think this is at the point where I've done what I can for the first pass.

Sounds good.  I will poke at it some more. 

> namely the nbootstraps which was adding a slow error computation to the KDE and Spline methods.

KDE is pretty fast, but spline is quite slow.  I'll look over it.  Again, the kluge is to use the histogram uncertainties, since most of the uncertainty is the data, not the method.  Since we we are generally just looking for LARGE deviations, getting the uncertainties off by 50-100% will not hurt. 

But why don't I evaluate this first, play around with it, and see what I think.

setup.cfg
```diff
@@ -8,4 +8,9 @@ omit =
 description-file = README.md
 
 [aliases]
-test = pytest
\ No newline at end of file
+test = pytest
+
+[tool:pytest]
+filterwarnings =
+    ignore::DeprecationWarning:patsy.*
```
jaimergp: Nice!

Issue #341, opened 2020-02-16
Address memory issues


mrshirts:
When running biggish systems (500 states, 2000 samples/state), it takes between ~11 and 20 GB or so (on OS X), meaning that on a 16GB computer, the time limiting factor is memory swaps, not computation.  So we should spend some time figuring out how to minimize memory usage.

jaimergp:
Some ideas for memory profiling [here](https://www.pluralsight.com/blog/tutorials/how-to-profile-memory-usage-in-python).

mrshirts:
Started going through memory profiling.  Once using mprofile, I identified approximately where the issues were, and then just stepped through with PDB to discover the exact commands.

Fundamentally, there are large data structures that are NxK large that are very hard to avoid.  We need to know the u_kn matrix to do anything.  We need to keep track of the W_kn weights.

* For the u_kn matrices, fundamentally, we need the data.  However, we might want to have the option to have u_kn be a callable function.  Exactly how to implement it will depend on the situation. For example, if the k states are K temperatures, then it's very easy to implement such a function. But if you need to evaluate an alchemical potential it could be complicated.  And accelerating such a function within the context of the mbar evaluations would be nontrivial. 

* The formula for the w_kn are w_kn = exp(-u_k(x_n)) / \sum k N_k exp(f-u_k(x_n).  It seems like it might be possible to just store the Nx1 denominators d(x_n), and reconstruct the NxK entries of the matrix when needed in the calculation, without creating the full matrix. We should probably store the log weights, for underflow reasons, but then in that case \log w_kn = -u_k(x_n) - \log d(x_n), and we can easily keep the simple sum, which with accelerated computations should be cheap.  However, rewriting this touches some of the most compute time-sensitive parts of the code, so should probably be done in conjunction with computational acceleration.

* It does seem that at minimum we should not have more than one u_kn and one w_kn matrix stored, and maybe just one u_kn. So we should see if we could do some calculations in place. Any thoughts appreciated!

* However, there are a number of things that could potentially be memory optimized beyond "keep one w_kn and one u_kn).   I'll start keeping track of those in extra comments.

mrshirts:
OK, so one thing that surprises me that takes extra memory: in precondition_u_kn(u_kn, N_k, f_k), the command:

   u_kn = u_kn - u_kn.min(0)

[...]

PR #359, opened 2020-02-23
Solved a number of issues that avoid lost memory

mrshirts:
A few changes for better memory management. In very large (500 states x 1000 samples/state) case, reduced memory by about 1/2. Specific changes detailed in issue #341, but in general, removed a number of cases where arrays were duplicated needlessly.

### DIFF
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -50,7 +50,6 @@ DEFAULT_SOLVER_PROTOCOL = mbar_solvers.DEFAULT_SOLVER_PROTOCOL
 # MBAR class definition
 # =========================================================================
 
-
 class MBAR:
     """
 
@@ -193,7 +192,7 @@ class MBAR:
             u_kn = kln_to_kn(u_kn, N_k=self.N_k)
 
         # u_kn[k,n] is the reduced potential energy of sample n evaluated at state k
-        self.u_kn = np.array(u_kn, dtype=np.float64)
+        self.u_kn = np.array(u_kn, dtype=np.float64, copy=False)
 
         K, N = np.shape(u_kn)
 
--- a/pymbar/mbar_solvers.py
+++ b/pymbar/mbar_solvers.py
@@ -68,10 +68,9 @@ def self_consistent_update(u_kn, N_k, f_k):
 
     u_kn, N_k, f_k = validate_inputs(u_kn, N_k, f_k)
     
-    states_with_samples = (N_k > 0)
-
+    u_kn_nonzero, N_k_nonzero, f_k_nonzero, states_with_samples = get_nonzero_states(u_kn, N_k, f_k)
     # Only the states with samples can contribute to the denominator term.
-    log_denominator_n = logsumexp(f_k[states_with_samples] - u_kn[states_with_samples].T, b=N_k[states_with_samples], axis=1)
+    log_denominator_n = logsumexp(f_k_nonzero - u_kn_nonzero.T, b=N_k_nonzero, axis=1)
     
     # All states can contribute to the numerator term.
     return -1. * logsumexp(-log_denominator_n - u_kn, axis=1)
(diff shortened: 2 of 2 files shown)


mrshirts:
Huh, I'm not sure what the issues is. My first guess was that the data was being modified some way by mbar, since now the links are better.  But doesn't look like it.

To tired tonight to finish debugging . . .

mrshirts:
Well, it looks like the issue was the change: 

Works:

`u_kn = u_kn - u_kn.min(axis=0)`

Fails

`u_kn -= u_kn.min(axis=0)` 

I don't get it. Must have something to do with modifying the original array.  We shouln't be changing the free energies, but could change observables.  So I guess we shouldn't do that? I can see why that's an issue, if we use the SAME internal array later to compute something else.  But I didn't see a difference when I checked?  I'll have to check again.  Will figure out later.

mrshirts:
Still need to debug a bit more, but IF we need to copy that array (and if we do, it should be done more explicitly) to preserve the inputs, we should perhaps need to consider preconditioning inside the any acceleration loops, instead of outside.

Probably the place to do it is when creating the u_kn_nonzero array, as if it is a copy, then it doesn't matter if it's altered.

Lnaden:
I figured out why this is failing, not sure on a fix yet. 

Under this scheme, doing the `-=` operator on the call `u_kn -= u_kn.min(0)` does correctly prevent a copy, but in doing so changes the value of that array everywhere. Given the following back trace:

```
-> self.f_k = mbar_solvers.solve_mbar_for_all_states(self.u_kn, self.N_k, self.f_k, solver_protocol)
  mbar_solvers.py(556)solve_mbar_for_all_states()
-> f_k_nonzero, all_results = solve_mbar(u_kn_nonzero, N_k_nonzero,
  mbar_solvers.py(524)solve_mbar()
-> f_k_nonzero, results = solve_mbar_once(u_kn_nonzero, N_k_nonzero, f_k_nonzero, **options)
  mbar_solvers.py(436)solve_mbar_once()
-> u_kn_nonzero = precondition_u_kn(u_kn_nonzero, N_k_nonzero, f_k_nonzero)
> mbar_solvers.py(386)precondition_u_kn()
-> u_kn -= u_kn.min(axis=0)  # do not use u_kn = u_kn + X, creates new u_kn array.
```

The following pointers then exist:
1. `u_kn -= u_kn.min(axis=0) ` references the `u_kn` arg of `precondition_u_kn`
2. `precondition_u_kn` accepts the `u_kn_nonzero` reference for its `u_kn` arg
3. This is the same `u_kn_nonzero` which has been passed into `solve_mbar_once` and `solve_mbar` above it.
4. This `u_kn_nonzero` is a sliced *view* generated from `get_nonzero_states` which accepted `u_kn` 
5. This `u_kn` was passed from `mbar_solvers.solve_mbar_for_all_states(self.u_kn, ...)` in the main `MBAR` class.

Since the `u_kn` of `precondition_u_kn` function is really just a pointer to a *view* of the `self.u_kn` matrix, doing `-=` results in the `self.u_kn` matrix being changed at those values, hence why the tests fail. I suspect having the `copy=False` kwarg in the `__init__` could do very similarly unexpected things if people modify the input `u_kn` from outside the MBAR class, but I haven't tested that yet. 

I don't have a solution for this, but at least that is what is happening here.

mrshirts:
Yes, that's the conclusion I came to, I should have documented that here, but didn't think that anyone else was looking at this issue currently. 

We do need at least two copies of u_kn - one that is passed into MBAR, and one that is internal to MBAR.  There are currently (in pymbar4) 3 copies, and I can cut it down to two (instead of 1, in this PR - the one passed in and the u_kn_nonzero array), but will need to tweak things a bit.  Was just waiting until the reformatting was done to implement that.

mrshirts:
Closing for now, since these changes can't be done without breaking some applications, but keeping referenced in #341 which will be left open for now to document the changes.

PR #360, opened 2020-02-24
From print to logging

jaimergp:
Converted `print` calls to `logging`.

Criteria for the levels:

- By default, they use `logger.info`.
- If the message was prefixed with `Warning` or was part of a check, it uses `logger.warning`.
- If the `print` call is part of a `if verbose:` block, they use `logger.debug`.

## Todo

- [ ] Remove `if verbose` lines and replace with a `logger.setLevel` call.
- [ ] Provide handlers and formatters as necessary.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "python.linting.enabled": true,
+    "python.linting.pylintEnabled": true,
+    "python.pythonPath": "/home/jaime/.conda/envs/pymbar/bin/python"
+}
--- a/pymbar/bar.py
+++ b/pymbar/bar.py
@@ -46,10 +46,15 @@ __license__ = "MIT"
 #=============================================================================================
 # IMPORTS
 #=============================================================================================
+import logging
 import numpy as np
 import numpy.linalg
 from pymbar.utils import ParameterError, ConvergenceError, BoundsError, logsumexp
 from pymbar.exp import EXP
+
+
+logger = logging.getLogger(__name__)
+
 
 def BARzero(w_F, w_R, DeltaF):
     """A function that when zeroed is equivalent to the solution of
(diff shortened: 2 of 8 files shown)


mrshirts:
Why does CI fail here? Simply because so much of the code changes?

jaimergp:
I don't fully understand the difference between `codecov/{patch,projecŧ}`:

From the [codecov docs](https://docs.codecov.io/docs/commit-status):

> The codecov/project status measures overall project coverage and compares it against the base of the pull request or parent commit.

> The codecov/patch status only measures lines adjusted in the pull request or single commit, if the commit is not in a pull request. This status provides an indication on how well the pull request is tested.

Maybe [this example](https://docs.codecov.io/docs/commit-status#section-patch-status) is useful. What I understand from that description is that I must have introduced changes that are not being tested (?), but I can't honestly see why or how.

mrshirts:
What's the difference between .info and .debug?  I think that in general, the verbosity results can be fairly useful, so if people see them as "debugging" they might not look at the them much.  I would rather have a lot of information be on, with the capability to turn it off, than the other way around.  

Of course, that might simply be a matter of setting the logger state to have debug = True.

jaimergp:
Logging works by setting message "levels". These are [the default values](https://docs.python.org/3/library/logging.html#logging-levels):

```
CRITICAL 50
ERROR 40
WARNING 30
INFO 20
DEBUG 10
NOTSET 0
```

Libraries only "emit" messages with a level label. It's up to applications to set the threshold level to "publish" (write to stdout, log file, whatever) these messages. Recommended behavior is usually warning and above. When verbosity is increased, we lower the threshold to info and above. Even more verbosity can be achieved by lowering it to debug and above.

We need to decide how important these messages and label them accordingly, but I am not familiar enough with the code to infer this.

mrshirts:
Got it!  I can todo that task, but it might take a bit of time.   

Hmm. Maybe the fastest way forward is to set everything that was sent directly to stdout w/o any error messaging to logging.info, and then add an issue for me to go back and adjust appropriately.

jaimergp:
By "without any error" you also mean "warnings"?

mrshirts:
> By "without any error" you also mean "warnings"?

Yes, exactly this.

mrshirts:
I'll go ahead and merge.

Issue #362, opened 2020-02-24
How do we approach the linting checks


jaimergp:
I ran `pylint` on `pymbar` so we only get warnings and errors (omitted style lint and the output from `old_mbar.py` and `tests`).

```
pylint --disable=all --enable=W,E pymbar/ --disable=W0511
```

<details>

<summary>This is the output</summary>

```
************* Module pymbar
pymbar/__init__.py:31:0: W1618: import missing `from __future__ import absolute_import` (no-absolute-import)
pymbar/__init__.py:40:0: W0702: No exception type(s) specified (bare-except)
pymbar/__init__.py:39:4: W0406: Module import itself (import-self)
************* Module pymbar.utils
pymbar/utils.py:26:0: W1618: import missing `from __future__ import absolute_import` (no-absolute-import)
pymbar/utils.py:374:4: W0107: Unnecessary pass statement (unnecessary-pass)
pymbar/utils.py:383:4: W0107: Unnecessary pass statement (unnecessary-pass)
pymbar/utils.py:392:4: W0107: Unnecessary pass statement (unnecessary-pass)
pymbar/utils.py:401:4: W0107: Unnecessary pass statement (unnecessary-pass)
************* Module pymbar.mbar
pymbar/mbar.py:40:0: W1618: import missing `from __future__ import absolute_import` (no-absolute-import)
pymbar/mbar.py:418:100: W1619: division w/o __future__ statement (old-division)
pymbar/mbar.py:467:20: W0612: Unused variable 'eigevec' (unused-variable)
pymbar/mbar.py:564:33: W0613: Unused argument 'warning_cutoff' (unused-argument)
pymbar/mbar.py:982:8: W0612: Unused variable 'N' (unused-variable)
pymbar/mbar.py:1093:8: W0612: Unused variable 'K' (unused-variable)
pymbar/mbar.py:1113:8: W0612: Unused variable 'expectations' (unused-variable)
pymbar/mbar.py:1113:22: W0612: Unused variable 'uncertainties' (unused-variable)
pymbar/mbar.py:1113:37: W0612: Unused variable 'covariances' (unused-variable)
pymbar/mbar.py:1190:8: W0612: Unused variable 'Deltaf_ij' (unused-variable)
pymbar/mbar.py:1190:19: W0612: Unused variable 'dDeltaf_ij' (unused-variable)
pymbar/mbar.py:1258:11: W0612: Unused variable 'N' (unused-variable)
[...]

PR #367, opened 2020-02-28
Add linting to CI

jaimergp:
Using `pylint` in a separate job _outside_ the build matrix, since this is just static code analysis.


mrshirts:
Just checking, what report should I be looking at to fix things?  The output  listed in the initial issue statement for https://github.com/choderalab/pymbar/issues/362? If so, which hash was that done off of to make sure the line # are right?

jaimergp:
* Linting report: https://github.com/choderalab/pymbar/pull/367/checks?check_run_id=475015683#step:4:10
* Commit: 622ed57cb95c3a6bbd2bac576fa25737c59b6df3 (this PR)

jaimergp:
Cleaned some unneeded warnings. Use this log now:
https://github.com/choderalab/pymbar/pull/367/checks?check_run_id=480428657#step:4:10

Now it's way less daunting!

mrshirts:
pymbar/__init__.py:40: [W0702(bare-except), ] No exception type(s) specified
    pymbar/__init__.py:39: [W0406(import-self), ] Module import itself

Not sure the right fix for these above.

    pymbar/pmf.py:1867: [W0640(cell-var-from-loop), PMF._bspline_calculate_h.ddexpf] Cell variable i defined in loop
    pymbar/pmf.py:1867: [W0640(cell-var-from-loop), PMF._bspline_calculate_h.ddexpf] Cell variable j defined in loop

Still figuring out what to do above.

    pymbar/pmf.py:756: [E0203(access-member-before-definition), PMF.generatePMF] Access to member 'kde' before its definition line 762

Not sure I agree here?  

    pymbar/pmf.py:1951: [W0105(pointless-string-statement), ] String statement has no effect

Yes, it's notes for now.

Still looking at a few more.

pymbar/mbar.py
```diff
@@ -307,7 +305,7 @@ def __init__(self, u_kn, N_k, maximum_iterations=10000, relative_tolerance=1.0e-
             self._initializeFreeEnergies(verbose, method=initialize)
 
             if self.verbose:
-                logger.info("Initial dimensionless free energies with method {}".format(initialize))
+                logger.info("Initial dimensionless free energies with method {:s}",format(initialize))
```
jaimergp: This should be `.format` (notice the comma). Also `{:s}` is totally equivalent to `{}`, so I don't think we should go this way. I think what `pylint` was trying to say is that we should use old-style formatting (`%` based) for the logging calls!

I hadn't realized this but some there are _some_ good points which might or might not be irrelevant to us (taken from [here](https://www.reddit.com/r/Python/comments/387oog/when_will_logging_use_newstyle_string_formatting/crsz00v/)):

- Since the arguments are separate from the template, the `logging` module can choose when to apply the string interpolation/replacement saving some CPU cycles. This is, the message will only be formatted if the logging level is sufficient.
- Having the template available means that logging aggregators (like `sentry`) can categorize log messages way better.

So in other words, `pylint` is trying to say that we should write it this way:


```suggestion
                logger.info("Initial dimensionless free energies with method %s", initialize)
```

jaimergp:
Thanks for the commits! I will reply below:

1) These warnings:

```
pymbar/__init__.py:40: [W0702(bare-except), ] No exception type(s) specified
pymbar/__init__.py:39: [W0406(import-self), ] Module import itself
```

... come from these lines in `pymbar.__init__`

```
try:
    from pymbar import version
except:
    # Fill in information manually.
    # TODO: See if we can at least get the git revision info in here.
    version = 'dev'
    full_version = 'dev'
    git_revision = 'dev'
    isrelease = False
```

The quickest fix would be using `except ImportError:`, but I don't see a `pymbar.version` module so this is always going to be the case (so this exception is not exceptional anymore). That said, we should be using `versioneer` for this kind of things.

2) It refers to this block (starts at line 1858)

```
            for i in range(nspline - 1):
                for j in range(0, i + 1):
                    if np.abs(i - j) <= kdegree:
                        def ddexpf(x, k): return db_c[i + 1](x) * db_c[j + 1](x) * expf[k](x)
``` 

`ddexpf` is using `i` and `j` from the outer scope, which will only work if `ddexpf` is only used inside that scope.  We are also redefining a function several times in that tight loop, which might be inefficient? Ideally, `ddexpf` should be defined outside the loop and take `i` and `j` as arguments too. Same should be applied in a similar block starting at line 1868.

3) Concerns this block:

```
            elif pmf_type == 'kde':

                # reshape data if needed.
                # it's a 1D array, instead of a Nx1 array.  Reshape.
                if len(np.shape(x_nb)) == 1:
                    x_nb = x_nb.reshape(-1, 1)

                if b > 0:
                    kde = KernelDensity()  # need to create a new one so won't get refit
                    params = self.kde.get_params()
                    kde.set_params(**params)

                kde.fit(x_nb, sample_weight=self.w_n)

                if b == 0:
                    self.kde = kde
                else:
                    self.kdes.append(kde)
```

`kde` won't be defined if `b <= 0`. By the way, `generatePMF` is a huge function that could benefit from breaking it into smaller pieces. Basically, each `pmf_type` could be a separate function that gets called from `generatePMF` if needed.

pymbar/mbar.py
jaimergp: Oh, there's a compromise: http://plumberjack.blogspot.com/2010/10/supporting-alternative-formatting.html

jaimergp: Oh, apparently, there's a compromise. We can use new-style formatting with `logging` without losing the performance and aggregation abilities following [these instructions](https://stackoverflow.com/a/24683360).

jaimergp:
I am going to add some changes here, if that's ok with everybody:

- [ ] fix some more CI issues
- [ ] apply black on this branch

Then I'll open another PR to apply the `snake_case` names on `pymbar4`.

jaimergp:
Ok, MacOS CI had broken but it was an easy fix (permissions thingies). 

I have fixed the obvious stuff and plan to merge later unless somebody states otherwise. Standing issues:

```
************* Module pymbar.pmf
pymbar/pmf.py:1861: [W0640(cell-var-from-loop), PMF._bspline_calculate_h.ddexpf] Cell variable i defined in loop
pymbar/pmf.py:1861: [W0640(cell-var-from-loop), PMF._bspline_calculate_h.ddexpf] Cell variable j defined in loop
pymbar/pmf.py:753: [E0203(access-member-before-definition), PMF.generatePMF] Access to member 'kde' before its definition line 759
************* Module pymbar.old_mbar
pymbar/old_mbar.py:812: [W0702(bare-except), MBAR.computeExpectations] No exception type(s) specified
pymbar/old_mbar.py:1351: [W0702(bare-except), MBAR.computeEntropyAndEnthalpy] No exception type(s) specified
pymbar/old_mbar.py:1368: [W0702(bare-except), MBAR.computeEntropyAndEnthalpy] No exception type(s) specified
pymbar/old_mbar.py:2093: [E0401(import-error), MBAR._computeUnnormalizedLogWeights] Unable to import '_pymbar'
pymbar/old_mbar.py:2139: [W0702(bare-except), MBAR._computeUnnormalizedLogWeights] No exception type(s) specified
pymbar/old_mbar.py:2295: [W0105(pointless-string-statement), MBAR] String statement has no effect
pymbar/old_mbar.py:2373: [W0105(pointless-string-statement), MBAR._adaptive] String statement has no effect
pymbar/old_mbar.py:2423: [E1306(too-few-format-args), MBAR._adaptive] Not enough arguments for format string
************* Module pymbar.tests.test_pmf
pymbar/tests/test_pmf.py:7: [W0611(unused-import), ] Unused import sklearn
************* Module pymbar.tests.test_timeseries
pymbar/tests/test_timeseries.py:148: [W0105(pointless-string-statement), test_detectEquil_constant_trailing] String statement has no effect
pymbar/tests/test_timeseries.py:10: [W0611(unused-import), ] Unused statsmodels.api imported as sm
```

I am not too concerned about `old_mbar` and `tests/*`, but we should probably revisit issues in `pmf.py` at some point. I think @mrshirts has created an issue about it here #369, so that's fine!

mrshirts:
> I am not too concerned about old_mbar and tests/*, but we should probably revisit issues in pmf.py at some point. I think @mrshirts has created an issue about it here #369, so that's fine!

* Yes, we shouldn't worry about old_mbar, since that will be dropped just before pymbar4 release. 
* I'd like to think about pmf reorganization and those issuues more holistically, and yes that's filed as an issue. 
* The test import ones should be looked at at some point, but it's not vital; create an issue for them and move on for now (I don't know how to fix them myself).

PR #368, opened 2020-03-05
Add deprecation warnings for camelCase names and add snake_case variants

jaimergp:
Addresses #365 

I added a temporary `_deprecate` function in `_deprecate.py` to ease the transition and avoid duplication. I have tested it with functions and should also work with methods...

I took `master` as a base, but most of the changes here won't reach Pymbar 4 (only the name replacements), and we have this excellent table here now:

```
BAR_zero                                             BARzero
BAR_zero                                             computeBARzero
BAR                                                  computeBAR
order_replicates                                     OrderReplicates
anderson_darling                                     AndersonDarling
QQ_plot                                              QQPlot
generate_confidence_intervals                        generateConfidenceIntervals
EXP_gauss                                            EXPGauss
EXP                                                  computeEXP
EXPGauss                                             computeEXPGauss
weights                                              getWeights
compute_effective_sample_number                      computeEffectiveSampleNumber
compute_overlap                                      computeOverlap
compute_free_energy_differences                      getFreeEnergyDifferences
compute_expectations_inner                           computeExpectationsInner
compute_covariance_of_sums                           computeCovarianceOfSums
compute_expectations                                 computeExpectations
compute_multiple_expectations                        computeMultipleExpectations
compute_perturbed_free_energies                      computePerturbedFreeEnergies
compute_entropy_and_enthalpy                         computeEntropyAndEnthalpy
compute_pmf                                          computePMF
statistical_inefficiency                             statisticalInefficiency
statistical_inefficiency_multiple                    statisticalInefficiencyMultiple
integrated_autocorrelation_time                      integratedAutocorrelationTime
integrated_autocorrelation_time_multiple             integratedAutocorrelationTimeMultiple
normalized_fluctuation_correlation_function          normalizedFluctuationCorrelationFunction
normalized_fluctuation_correlation_function_multiple normalizedFluctuationCorrelationFunctionMultiple
subsample_correlated_data                            subsampleCorrelatedData
detect_equilibration                                 detectEquilibration
statistical_inefficiency_fft                         statisticalInefficiency_fft
detect_equilibration_binary_search                   detectEquilibration_binary_search
```

I have only modified callables for now. Attributes (if relevant), parameters and free variables are untouched for now.

PR #372, opened 2020-03-19
camelCase -> snake_case

jaimergp:
Following PEP8 recommendations and using table of replacements as provided in #368.

Note that `examples/` is not affected in this PR. They will be "modernized" in a separate PR.

## Questions

- [x] Shall I apply `snake_case` to public attributes and module-level variables?
- [x] Shall I apply `snake_case` to unexposed variables?

## Notes

* This PR introduces breaking changes! Check replacements in #368.
    * Special attention: `getFreeEnergyDifferences` is now `compute_free_energy_differences`, for consistency with other methods and functions.
* `pymbar.bar` and `pymbar.exp` _modules_ have been merged into `pymbar.other_estimators`. 

Summary of changed names:

```
AndersonDarling                                    anderson_darling
BARzero                                            bar_zero
computeBAR                                         bar
computeBARzero                                     bar_zero
computeCovarianceOfSums                            compute_covariance_of_sums
computeEffectiveSampleNumber                       compute_effective_sample_number
computeEntropyAndEnthalpy                          compute_entropy_and_enthalpy
computeEXP                                         exp
computeExpectations                                compute_expectations
computeExpectationsInner                           compute_expectations_inner
computeEXPGauss                                    exp_gauss
computeMultipleExpectations                        compute_multiple_expectations
computeOverlap                                     compute_overlap
computePerturbedFreeEnergies                       compute_perturbed_free_energies
computePMF                                         compute_pmf
detectEquilibration                                detect_equilibration
detectEquilibration_binary_search                  detect_equilibration_binary_search
EXPGauss                                           exp_gauss
generateConfidenceIntervals                        generate_confidence_intervals
generatePMF                                        generate_pmf
GetAnalytical                                      get_analytical
getFreeEnergyDifferences                           compute_free_energy_differences
getInformationCriteria                             get_information_criteria
getKDE                                             get_kde
getMBAR                                            get_mbar
getPMF                                             get_pmf
getWeights                                         weights
integratedAutocorrelationTime                      integrated_autocorrelation_time
integratedAutocorrelationTimeMultiple              integrated_autocorrelation_time_multiple
normalizedFluctuationCorrelationFunction           normalized_fluctuation_correlation_function
normalizedFluctuationCorrelationFunctionMultiple   normalized_fluctuation_correlation_function_multiple
OrderReplicates                                    order_replicates
PrintResults                                       print_results
QQPlot                                             qq_plot
statisticalInefficiency                            statistical_inefficiency
statisticalInefficiency_fft                        statistical_inefficiency_fft
statisticalInefficiencyMultiple                    statistical_inefficiency_multiple
subsampleCorrelatedData                            subsample_correlated_data
```


mrshirts:
If if is best practice to snake_case public attributes, module-level variables, and unexposed variables, now it is the time to do.

mrshirts:
Should we change things like EXP_gauss, BAR_gauss to exp_gauss and bar_gauss to be more snake_casey?

jaimergp:
All in for snake_casiness!

jaimergp:
There's a subtle problem:

If we go all lowercase, function `BAR` in `pymbar.bar` becomes `bar`. If we import that one in `pymbar.__init__`, it shadows `pymbar.bar` (module) to become `pymbar.bar` (function). Basically we can't have a function named as a module if we are going to expose that at a package level.

So, possible solutions:
* Keep the function named `BAR`, but it will be inconsistent with `bar_zero`.
* Rename the _function_ as `bennett_acceptance_ratio`. This must be done for similar acronyms (MBAR, PMF, etc).
* Rename the _modules_ so the acronyms are expanded.

Lnaden:
So PEP8 says the following:

> Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

But I think that's for Classes more generally. I think for the snake case, it would still be `bar` since they are functions. `MBAR` would still be a thing since its a class. But is it really so bad to have `from pymbar.bar import bar`, especially since we mask the `pymbar.bar` module with `__all__` anyways?

jaimergp:
It can lead to an inconsistency (I think that's why the tests are failing), where the _module_ `pymbar.bar` is unreachable. 

You are right about classes, so this only affects `BAR/bar` and `EXP/exp`.

Lnaden:
How frequently do we think people manually import the `bar` and `exp` modules?

jaimergp:
Not often, but the thing is that I think we literally make all the symbols in `pymbar.bar.*` unreachable this way?

jaimergp:
The only way we can get there is by importing `pymbar.bar` directly, but before importing the bare `pymbar`, maybe?

```
import pymbar.bar
import pymbar
```

Lnaden:
Well the only thing in the `bar` module right now to import are the BAR, BARZero (aside from the other imports), there is the logger for that module too, but thats it.

Lnaden:
> The only way we can get there is by importing pymbar.bar directly, but before importing the bare pymbar, maybe?

I'm against that, lets keep it simple, I think if there are limited imports anyways we should just rename the modules (or put them in a new module, `other_estimators`?) and then import the relevant methods.

jaimergp:
I am slightly inclined to rename modules rather than functions too. So let's see, would this work?

* bar -> bennett_acceptance_ratio
* exp -> exponential_averaging
* mbar -> multistate_bennett_acceptance_ratio
* mbar_solvers -> multistate_bennett_acceptance_ratio_solvers or simply solvers
* pmf -> potentials_of_mean_force

Lnaden:
I would argue in favor of keeping mbar, mbar_solvers, and pmf as is. Its what people mostly will want the mbar repo for and it really is only bar and exp which are the problems. It may be better to move those two into their own module if we are looking for consistency.

mrshirts:
> It may be better to move those two into their own module if we are looking for consistency.

I'd favor this.   other_estimators works for me.

jaimergp:
Ok, sounds fair! To be clear, what option are we agreeing on?

* Option A: Merging `pymbar/bar.py` and `pymbar/exp.py` into a new file called `pymbar/other_estimators.py` 
* Option B: Creating a subpackage `pymbar/other_estimators/` that will contain `bar.py` and `exp.py`.

Lnaden:
My vote is A, I like to avoid Subpackages when possible

mrshirts:
Ditto, A.

jaimergp:
Ok, that seems to do it. Please review whenever you have time.

I have dropped a couple of `TODO`s here and there regarding docstrings, too.

pymbar/tests/test_exp.py
```diff
@@ -4,7 +4,7 @@
 
 import numpy as np
 import pytest
-from pymbar import exp as pyexp
+from pymbar import other_estimators as other
```
mrshirts: I would probably do this as `import other_estimators as estimators`, since bar is an estimator, not an `other`.  Will make usage clearer.

pymbar/tests/test_bar.py
```diff
@@ -4,7 +4,7 @@
 
 import numpy as np
 import pytest
-from pymbar import bar as pybar
+from pymbar import other_estimators as other
```
mrshirts: Same as below (other -> estimator)

pymbar/timeseries.py
```diff
@@ -972,4 +972,4 @@ def detect_equilibration_binary_search(A_t, bs_nodes=10):
             start = time_grid[k - 1]
             end = time_grid[k + 1]
 
-    return (t, g, Neff_max)
+    return t, g, Neff_max
```
mrshirts: Remind me why this change? (Just so I know good formatting practices)

jaimergp: They are redundant in this case. You do need to add them for single-element tuple for extra clarity (see [here](https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas)), but for single expression statements like a `return`, they can (and IMO should) be omitted for n-element tuples.

jaimergp: There might be more, now that you bring this up.

jaimergp: Yep, four more. All of them in `timeseries`. Do you want me to replace those too?

mrshirts: Sure, go for it!

mrshirts:
I am OK with this being merged - not sure @jaimergp was waiting on anything else to merge.

PR #375, opened 2020-03-26
Make Pylint less noisy

### LINKED ISSUE #370: Revisit linting in tests
As per #367:

```
************* Module pymbar.tests.test_pmf
pymbar/tests/test_pmf.py:7: [W0611(unused-import), ] Unused import sklearn
************* Module pymbar.tests.test_timeseries
pymbar/tests/test_timeseries.py:148: [W0105(pointless-string-statement), test_detectEquil_constant_trailing] String statement has no effect
pymbar/tests/test_timeseries.py:10: [W0611(unused-import), ] Unused statsmodels.api imported as sm
```

Lnaden:
Adds some pylint ignore things to expected/canary behavior.
Fixing the PMF module will take a bit more refactoring, but its not
major for now.

Fixes #370

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -82,7 +82,7 @@ jobs:
     env:
       CI_OS: ubuntu-latest
       PACKAGE: "pymbar"
-      PYVER: "3.6"
+      PYVER: "3.7"
 
     steps:
     - name: Checkout the code
@@ -103,6 +103,8 @@ jobs:
         . devtools/github-actions/initialize_conda.sh
         conda activate test
         pylint $PACKAGE/
+      # Allow the lint to fail and not stop the other jobs
+      continue-on-error: true
 
     # TODO: Remove --exclude pymbar/old_mbar.py after deprecation
     - name: Run black check
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -103,8 +103,6 @@ jobs:
         . devtools/github-actions/initialize_conda.sh
         conda activate test
         pylint $PACKAGE/
-      # Allow the lint to fail and not stop the other jobs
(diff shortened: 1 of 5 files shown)


jaimergp:
For `black` you'll need to run `black -l 99` on `pmf.py`

jaimergp:
Btw, we have [this line](https://github.com/choderalab/pymbar/blob/f8496cfb30743afbc96a9c61b5992618d4001be9/.github/workflows/CI.yaml#L112) so `black` always runs!

Lnaden:
Don't you already run `black` with the `-l 99` flags on the whole package anyways? This is only focusing on the lint things.

I also made it so the pylint failures don't bring down the whole job in the CI.yaml file

jaimergp:
`black` must be run locally before submission. The CI only checks the output wouldn't be different. We can set up a bot to do it for us, but I haven't done that yet.

Lnaden:
> Btw, we have this line so black always runs!

Yes, I saw that, but I didn't want pylint failures to hold up a PR per-se. Black I get due to the nature of the formatting changes (and how large they can be), but pylint is more forgivable in my book. We can debate that point if you want though.

jaimergp:
Hm, the problem with adding that line is that GitHub will report it as successful (instead of ignored-failure, but that feature is not yet implemented in GHA, I think), so it's a bit misleading. I'd rather have the eyesore of a red cross in the CI checks and merge knowing that that error is due to acceptable lint errors. Otherwise we might forget to check if there are new linting issues when we forget about this decision.

jaimergp:
This thread addresses the topic: https://github.community/t5/GitHub-Actions/continue-on-error-allow-failure-UI-indication/td-p/37033

Lnaden:
Fair point, you convinced me otherwise.

PR #385, opened 2020-04-10
Refactoring pmf code

mrshirts:
First pass (still doesn't run) of restructuring the PMF code. 

Although it doesn't run, the revised structure should be pretty much in place.


pymbar/pmf.py
```diff
@@ -454,13 +454,19 @@ def generate_pmf(
 
         return result_vals  # should we return results under some other conditions?
 
-    def _setup_pmf_histogram(histogram_parameters):
+    def _setup_pmf_hiostogram(histogram_parameters):
```
jaimergp, line 457: Methods require a first argument `self` _or_ decorating them with `@staticmethod` if they don't need access to `self`.

```suggestion
    def _setup_pmf_histogram(self, histogram_parameters):
```
or 

```suggestion
    @staticmethod
    def _setup_pmf_histogram(histogram_parameters):
```

In this case you do need `self`, so we can only choose option A.

pymbar/pmf.py
```diff
@@ -454,13 +454,19 @@ def generate_pmf(
...
 
         if "bin_edges" not in histogram_parameters:
             raise ParameterError(
                 "histogram_parameters['bin_edges'] cannot be undefined with pmf_type = histogram"
                 )
 
+        self.histogram_parametes = histogram_parameters
```
jaimergp, line 464:
```suggestion
        self.histogram_parameters = histogram_parameters
```

pymbar/pmf.py
mrshirts: I think I addressed this in most recent commit - the first one was just spewing text up.  I'll go through and look for ones that don't access self and add the decorator.

pymbar/pmf.py
mrshirts: Spelling error fixed.

pymbar/pmf.py
mrshirts: And spelling error fixed.

mrshirts:
OK, some updates.  Passes all but three tests (involved in splines) now.

jaimergp:
Awesome! I added some changes!

* Updated examples to use the latest API changes.
* Fixed a couple of typos in `pmf.PMF`
* Ran black on needed files.
* There are some undefined variables in `pmf.PMF._get_initial_spline_points` and `pmf.PMF._generate_pmf_spline`. Namely: `nspline`, `kdegree`, `xrange`, `initivals`, and `fold`, `dx`, `xold`, respectively.

mrshirts:
Thanks!

> nspline, kdegree, xrange, 
Now properly defined, I guess the tests don't check this!

> initivals
No longer needed.

> fold, dx, xold

These all are defined before they are reached logically - the first time through the loop the conditional skipped over, and they are all defined after the conditional.

I merged and have cleaned up a bit more, doesn't quite pass all the tests for some reason, I will fix that before pushing it back up.

mrshirts:
Cleaned up some issues with the tests and spline codes.  Bootstrap uncertainties in histograms not quite right, will work on that.

mrshirts:
OK, most things are working.  There's still an issue with histogram bins numbering not being consistent among bootstraps in the histogram, which messes up the uncertainty there, so I will need to deal with that before merging.  That's the biggest remaining issue on this overall thread (probably still need to expand the pmf_test.py coverage), but that can/should be done separately. 

Oh, function docstrings.  I'll add that before final commits as well.  Anything else key to add?

jaimergp:
Mm, `pymbar/tests/test_pmf.py` was deleted as part of the last changes. Is that an accident?

mrshirts:
Whoops, wasn't supposed to have.  Added back in.

mrshirts:
pylint says there are three variables that are used before being set, but the logic of the code flow actually prevents that from happening. The place they are used is in a logical block that doesn't get called the first time through the loop.  

If there's a better way of writing the same logic flow, let me know!

mrshirts:
OK, looks like everything is working correctly now.  I will come back later this week and add some docstrings, but the logic should be in place.  

The pylint errors about variables being used before they are defined doesn't properly take program logic into account.

mrshirts:
OK, this should be ready to merge in mod any minor error caught.  PMF functionality broken up and data structures clarified somewhat, at least minimal docstrings on everything, black run. A couple bugs in MC functionality caught.

mrshirts:
The failing issue is that I renamed umbrella-sampling-smoothpmf.py to umbrella-sampling-advanced-pmf.py, but the example tests is still try to call umbrella-sampling-smoothpmf.py. I can't figure out where this specified, though - grepping through the pymbar directory, I can't figure out where that change should be made.  Where do I adjust that?

Lnaden:
Should be in the `.github/workflows/CI.yaml file` I think. It looks like it makes a direct call for python umbrella-sampling-smoothpmf.py

I can fix it real quick, I have it open

mrshirts:
Ah, grep -r doesn't match hidden files.  Thanks!

mrshirts:
OK, took a stab at fixing the black issues.

mrshirts:
Still failing lint because it thinks variable are accessed before defined, but it is misreading the program logic: those statements are bypassed the first time through the loop.

mrshirts:
OK, seems to be ready to go mod the pylint issues.

jaimergp:
I have added `# pylint: disable=XXX` comments where needed (block scoped) to get rid of the false positives there. I had to define `self.histogram_data` and `self.histogram_datas` in `__init__` for a cleaner fix, but I don't know if the presence of those attributes dictates any flow control later in the code. Tests seem to be passing so I guess it's not a problem to have them initialized to `None`. By the way, out of curiosity, why do we need both `*_data` and `*_datas`?

One thing not caught by pylint (but caught but the VS Code internal linter) is the presence of `initivals` in [this block](https://github.com/choderalab/pymbar/blob/2f5430ac478f6645dcc4a3b403fa5e09c695fb4d/pymbar/pmf.py#L852-L878). They don't seem to be defined anywhere before that, which might be an error, but I just woke up so maybe I am slow 😬

mrshirts:
> One thing not caught by pylint (but caught but the VS Code internal linter) is the presence of initivals in this block.

Yeah, that got eliminated somehow.  I searched back in the history and added where it was defined.

mrshirts:
>  had to define self.histogram_data and self.histogram_datas in __init__ for a cleaner fix, but I don't know if the presence of those attributes dictates any flow control later in the code.

I think that's fine.  Histogram_data stores the original results; histogram_datas stores the list of bootstrapped results.  I moved the definition down to setup_pmf_histogram, since it should not be called outside of that code path.

jaimergp:
Awesome, this looks ready to go, then!

mrshirts:
OK, I will go ahead and merge then!

Issue #380, opened 2020-03-28
better dimension handling of histograms in PMF


mrshirts:
Right now, the bin_edges expects an array of arrays.  If only 1D PMF, people should be able to just put in an array.

mrshirts:
Addressed by #387

mrshirts:
Addressed by #387

PR #387, opened 2020-04-24
Bin edges support

mrshirts:
Now support passing in a single array instead of a list of arrays for bin edges in 1D.


mrshirts:
Addresses #380

Issue #379, opened 2020-03-28
Check proper verbosity of mbar


mrshirts:
Doesn't appear to be printing out much anymore when verbose=True is set.

mrshirts:
OK, the issue is that setting verbose sends the desired text to logger.info, but doesn't actually print it. Presumably, if one sets verbose, one wants that info to be printed out in some way, but right now, it's getting lost.  @jaimergp , suggestions for sending that information to where the user can see it?

jaimergp:
The `logging` modules works on the assumption that _emission_ of messages is decoupled of _delivery_. In this decoupling, libraries are only emitters, and it's up to the applications to configure how to deliver the message. So the _user_ needs to configure `logging` to print stuff if they want to (usually, `logging.basicConfig()` is enough). 

Since `pymbar` is a library, does it make sense for us to configure the verbosity? Do we have an CLI entry point? We can configure `logging` globally with some settings (e.g. with a singleton like `pymbar.settings.verbose = True`, but in principle we _shouldn't_.

jaimergp:
That's why all this `if verbose:` checks do not make much sense in our case. We can (and should) leave all the `logger.{info,debug,warn,error}` calls unchecked and, if anything, configure the verbosity at a global level.

mrshirts:
So what would your suggestion be to have the code perform how people would expect, that if they give a verbosity argument, it increases the amount of information out? What hierarchy of flags to use?  Where and when to send the output out (right now, most of the output occurs near the data that is being generated).

jaimergp:
Ideally, there would be no `verbose` flag at the function/class level. Users are expected to set the logging level in their scripts if desired.

```
import logging
logging.basicConfig(level=logging.DEBUG)

# proceed with your pymbar calling code
```

To transition from the current approach, last version of pymbar 3 should include a deprecation warning for `verbose=True`:

```
>>> some_result = some_function(verbose=True)
DeprecationWarning("Pymbar v4.0+ will use `logging` for debug, informative reports and warnings. To increase output verbosity in the future, please use `logging.basicConfig(level=logging.<DESIREDLEVEL>)` in your scripts")
```

IIRC, there's a context manager you can use for temporary verbosity increase. The main idea is that emission/delivery are uncoupled in terms of responsibility, but it still happens synchronously if you want to. There's more info [here](https://realpython.com/python-logging/) (check the "Using handlers" section).

mrshirts:
[...]

PR #391, opened 2020-04-27
Configure logging with support for verbose flags

jaimergp:
Addresses #379

### DIFF
--- a/pymbar/__init__.py
+++ b/pymbar/__init__.py
@@ -28,6 +28,10 @@ __license__ = "MIT"
 __maintainer__ = "Levi N. Naden, Michael R. Shirts and John D. Chodera"
 __email__ = "levi.naden@choderalab.org,michael.shirts@colorado.edu,john.chodera@choderalab.org"
 
+from pymbar.utils import configure_logging
+
+configure_logging()
+
 from pymbar import timeseries, testsystems, confidenceintervals
 from pymbar.mbar import MBAR
 from pymbar.other_estimators import bar, bar_zero, exp, exp_gauss
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -49,6 +49,7 @@ from pymbar.utils import (
     DataError,
     logsumexp,
     check_w_normalized,
+    log_level,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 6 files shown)


jaimergp:
That's a good point. I think I can come up with something...

The `level` of the message is independent of the `verbose` flag. In other words, `verbose` does not control when a messages gets _logged_, but when the message is _emitted_ (in our case, to stdout by default). So what we are doing now is temporarily changing the emission threshold down from _warning_ to _info_. That's something we can't workaround because that's the backend for the `logging` mechanism.

However, I think we could wrap up this logic in a nicer way. Some quick ideas, let me know which one you like more:

```python
# A
with emit_if_verbose(__name__, verbose=verbose):
    logger.info("Something something")

# B
logger.log("Something something", level=logging.INFO, force_emit=verbose)
```

mrshirts:
I think I prefer the second? 

I think the second part is to document somewhere how people ask for emission of the other levels.

jaimergp:
Let me know what you think about it. It's a bit hacky, but hopefully not too much :)

My only concern is what happens when `pymbar` is used in another project that is already using `logging` without our custom subclass. Maybe that extra keyword can cause problems if it's not consumed before hand... but maybe the logging machinery is isolated enough? We'll see during the `rc` stage I guess.

mrshirts:
Anything else needed for this?

jaimergp:
Ideally, some manual testing, to see if the logging machinery behaves the way a user would expect. I did a couple of quick ones with the examples and was successful, but I'd like more feedback if possible.

jaimergp:
@mrshirts anything else you want to add here?

jaimergp:
This is needs some more testing in real world conditions, ideally in a notebook or somewhere where you are using other libraries at the same time. Does anybody have a good example?

Lnaden:
My one question with this, is there a reason you cannot add a different level for verbosity? Since `INFO`, `DEBUG`, etc are all just `int`, can we not make a level for `VERBOSE` and then if `verbose` we just set the logger level to that? I think that would be much simpler than carrying around the `force_emit=verbose` in all the calls.

mrshirts:
The one thing that I want to make sure of is it's very clear to users in the documentation how to make a given function call verbose.

PR #397, opened 2020-05-20
Replace "potential of mean force" with "free energy surface"

mrshirts:
Improve the terminology by replacing "potential of mean force" with "free energy surface" and "PMF" with "FES". I considered "free energy profile", but that would mean the abbreviation is FEP, which is already overloaded. I still use "free energy profile" in the text description when referring to 1D "surfaces".


mrshirts:
Cool, passed first time.  Let me know thoughts on change.

mrshirts:
@jchodera  let me know what you think!

mrshirts:
OK, no comments, so merging.

Issue #340, opened 2020-02-16
Accelerate the code, starting with the inner loops


mrshirts:
We need to accelerate the inner loops. 

I think the most important things to do would be to accelerate in this order

* logsumexp.  

This is the largest cos. 

After that is accelerated, it's probably worth it to look at:

* mbar_gradient - used the most in many algorithms, just a few lines wrapped around logsumexp
8 self_consistent_iteration - used a lot, also just a few lines.
* mbar_hessian - not many lines either, though also calls some other routines such as mbar_W_nk that are themselves relatively short wrappers around logsumexp.

jaimergp:
I'd strongly suggest running `python -m cprofile -o profile.dat whatever_pymbar_script.py` to profile all the calls and then visualize the output with `snakeviz`.

It will produce a visualization [similar to this one](https://github.com/choderalab/openmmtools/pull/434#issuecomment-516465268), which I found extremely informative to find the bottlenecks.

mrshirts:
Ah, I had been using cProfile, but was poring over the text: I didn't know about snakeviz.

Here's three sample profiles of a set of harmonic oscillators, using the adaptive method - all of them converged in 3 self-consistent and 3 Newton-Raphson iterations.

100 states, 25000 samples per state: [profile.100x25000.txt](https://github.com/choderalab/pymbar/files/4216685/profile.100x25000.txt) 
1000 states, 300 samples per state: [profile.1000x300.txt](https://github.com/choderalab/pymbar/files/4216686/profile.1000x300.txt)
500 states, 1000 samples per state: [profile.500x1000.txt](https://github.com/choderalab/pymbar/files/4216687/profile.500x1000.txt)

I'm not posting the graphs here, since you need to do a little interactive work to see what is going on, but you can visualize them easily with this. 

All of these took about 3-4 minutes. Note that anything bigger (size scales roughly (nsamples/state)*(n_state)^2 starts to cause memory issues, and start running much more slowly because of page faults. All of these ran at about 10G, and took about the same amount of time (1000x300 was slower, but bigger).  

Interestingly, I noticed it cycling back and forth between 100% and 400% CPU usage, so some aspects are well-parallelized, and some are not.

So what I identified as the slow parts were indeed, but the balance is not as clear.  

`logsumexp` is the most time-consuming part, but the non-`logsumexp` parts of `mbar_gradient`, `self_consistent_update`, and `mbar_hessian` were expensive as well if I'm interpreting it correctly. 

[...]

PR #400, opened 2020-06-29
[WIP] made whole adaptive workflow jax/jit

mrshirts:
Converted the entire loop in adaptive to jax/jit. 
See discussion on issue #340 for discussion of timings.

Intended for inspection, not merging for now. 

Failing because jax/jit isn't loaded via conda (not clear it can be right now, conda-forge install on my machine didn't work, had to pip install).

### DIFF
--- a/pymbar/mbar_solvers.py
+++ b/pymbar/mbar_solvers.py
@@ -45,7 +45,6 @@ def validate_inputs(u_kn, N_k, f_k):
     """
     n_states, n_samples = u_kn.shape
 
-    return u_kn, N_k, f_k
     u_kn = ensure_type(u_kn, 'float', 2, "u_kn or Q_kn", shape=(n_states, n_samples))
     N_k = ensure_type(N_k, 'float', 1, "N_k", shape=(n_states,), warn_on_cast=False)  # Autocast to float because will be eventually used in float calculations.
     f_k = ensure_type(f_k, 'float', 1, "f_k", shape=(n_states,))
@@ -54,15 +53,15 @@ def validate_inputs(u_kn, N_k, f_k):
 
 def jax_self_consistent_update(u_kn, N_k, f_k, states_with_samples=None):
 
+    jNk = 1.0*N_k
+
     # Only the states with samples can contribute to the denominator term.
     if states_with_samples is not None:
-        log_denominator_n = logsumexp(f_k[states_with_samples] - u_kn[states_with_samples].T, b=1.0*N_k[states_with_samples], axis=1)
+        log_denominator_n = logsumexp(f_k[states_with_samples] - u_kn[states_with_samples].T, b=jNk[states_with_samples], axis=1)
     else:
-        log_denominator_n = logsumexp(f_k - u_kn.T, b=1.0*N_k, axis=1) 
+        log_denominator_n = logsumexp(f_k - u_kn.T, b=jNk, axis=1) 
     # All states can contribute to the numerator term.
     return -1. * logsumexp(-log_denominator_n - u_kn, axis=1)  # check transpose
-
-jit_self_consistent_update = jax.jit(jax_self_consistent_update)
 
 def self_consistent_update(u_kn, N_k, f_k, states_with_samples=None):
     """Return an improved guess for the dimensionless free energies
@@ -85,9 +84,7 @@ def self_consistent_update(u_kn, N_k, f_k, states_with_samples=None):
     -----
     Equation C3 in MBAR JCP paper.
     """
-
-    u_kn, N_k, f_k = validate_inputs(u_kn, N_k, f_k)
-    return jit_self_consistent_update(u_kn, N_k, f_k, states_with_samples=states_with_samples)
+    return jax_self_consistent_update(u_kn, N_k, f_k, states_with_samples=states_with_samples)
 
 def jax_mbar_gradient(u_kn, N_k, f_k):
 
@@ -95,8 +92,6 @@ def jax_mbar_gradient(u_kn, N_k, f_k):
     log_denominator_n = logsumexp(f_k - u_kn.T, b=jNk, axis=1)
     log_numerator_k = logsumexp(-log_denominator_n - u_kn, axis=1)
     return -1 * jNk * (1.0 - jnp.exp(f_k + log_numerator_k))
-
-jit_mbar_gradient = jax.jit(jax_mbar_gradient)
 
 def mbar_gradient(u_kn, N_k, f_k):
     """Gradient of MBAR objective function.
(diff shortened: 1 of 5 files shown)


pymbar/mbar_solvers.py
```diff
@@ -3,7 +3,12 @@
 import numpy as np
 import math
 import scipy.optimize
-from pymbar.utils import ensure_type, logsumexp, check_w_normalized
+from pymbar.utils import ensure_type, check_w_normalized
+import jax
+from jax.scipy.special import logsumexp
+from jax.ops import index_update, index
+from jax.config import config; config.update("jax_enable_x64", True)
```
proteneer, line 10: This will slow things down dramatically, esp. on the GPU. Do you really need 64bit precision?

pymbar/mbar_solvers.py
```diff
@@ -289,105 +297,87 @@ def adaptive(u_kn, N_k, f_k, tol=1.0e-12, options=None):
...
+    results['f_k'] = f_k
+    results['max_delta'] = max_delta
+    results['iterations'] = iteration
+    results['gnorms_sci'] = gnorms_sci
+    results['gnorms_nr'] = gnorms_nr
+    results['which_method'] = which_method
+
+    return results
 
+jit_core_adaptive = jax.jit(core_adaptive,static_argnums=(0,1,2,3,4,5,))
```
proteneer, line 373: When you call jax.jit, the first invocation will be exceedingly slow it's compiling the JIT kernels. So if you're benchmarking you should call this *more* than once to amortize out the jit time.

Futhermore, if the arguments have shapes that contain (None,) in one of the shapes, a recompilation is required due to an XLA requirement for each specialization of (None,) into a known shape.

pymbar/mbar_solvers.py
```diff
@@ -289,105 +297,87 @@ def adaptive(u_kn, N_k, f_k, tol=1.0e-12, options=None):
...
+    nr_iter = 0
+    sci_iter = 0
 
-    for iteration in range(0, options["maximum_iterations"]):
+    which_method = []
+    gnorms_sci = []
+    gnorms_nr = []
 
+    # Perform Newton-Raphson iterations (with sci computed on the way)
+    for iteration in range(0, maximum_iterations):
```
proteneer, line 320: can you write this as jax.lax.while_loop? Just jit-compiling this for loop while be a nightmare for XLA

pymbar/mbar_solvers.py
mrshirts: Probably yes - the algorithm doesn't seem to converge well with floats.  I can poke around and see what the deal is.

pymbar/mbar_solvers.py
```diff
@@ -145,16 +149,28 @@ def mbar_objective_and_gradient(u_kn, N_k, f_k):
...
-    grad = -1 * N_k * (1.0 - np.exp(f_k + log_numerator_k))
+    grad = -1 * N_k * (1.0 - jnp.exp(f_k + log_numerator_k))
 
     obj = math.fsum(log_denominator_n) - N_k.dot(f_k)
 
     return obj, grad
 
+def jax_mbar_hessian(u_kn, N_k, f_k):
+
+    jNk = 1.0*N_k
```
proteneer, line 163: what's the point of 1.0*?

pymbar/mbar_solvers.py
mrshirts: Ah, this explains why the partial jitification works better- then it's calling different functions multiple times.  Though only 7-8 times in many cases.

mrshirts: I will test running mbar initialization several times with different data sets in the same script to see if that helps.

pymbar/mbar_solvers.py
mrshirts: > can you write this as jax.lax.while_loop?

I can take a look, though the fact that it's on the adaptive loop that is generally only called once in pymbar means it won't be as useful to do this.

pymbar/mbar_solvers.py
proteneer: I also noticed you're declaring every arg to be static, which means that unless you're passing in the exact args you'll be triggering a recompilation. Why did you have to declare them to be static to start with?

proteneer: U_kn probably shouldn't be static if you're trying to pass in results from different runs?

pymbar/mbar_solvers.py
mrshirts: N_k is an int, jit (or maybe Jax, I can't recall) complains if it's not converted to float first.  I wasn't able to find the best direct conversion to float function that made jit happy. so I just multiplied by 1.0 to do it automatically.

pymbar/mbar_solvers.py
proteneer: maybe it makes sense then to just jit the body of this function? (i.e. move everything inside the for loop to a separate function and just jit that)

mrshirts: Ah, yes, that makes a lot of sense given the constraints. I'll try that.

pymbar/mbar_solvers.py
mrshirts: I had to declare it static because otherwise, gnorm_sci and gnorm_nr end up as traced arrays instead of concrete arrays (since they are functions of traced arrays), and jit refuses to compile the conditional comparing the two, which decides the branch to take, unless they are declared static.

mrshirts: @proteneer had a good suggestion for handling the conditional:

import jax
import jax.numpy as jnp
# can't be JIT'd
def foo(a, b, cond_a, cond_b):
    if cond_a < cond_b:
        return a+b
    else:
        return a-b
# foo_jit = jax.jit(foo) # fails
print(foo(0., 1., 2., 3.))
# can be JIT'd
def bar(a, b, cond_a, cond_b):
    return jnp.where(cond_a < cond_b, a+b, a-b)
bar_jit = jax.jit(bar)
print(bar_jit(0., 1., 2., 3.))

pymbar/mbar_solvers.py
mrshirts: Still need to work on this . . .

pymbar/mbar_solvers.py
mrshirts: So, if we accelerate just the inner loop, then we can pull the conditional out of the accelerated code with little loss of timing - all of the conditionals are just comparing floats and assigning values, so there is little use in accelerating them.

mrshirts:
Closing in favor of one that just accelerates the inner loop.

Issue #396, opened 2020-05-14
conda install -c omnia pymbar fails for Python 3.7.4


vvoelz:
Hi John et al:

I'm getting an error when I try to install `pymbar` using `$ conda install -c omnia pymbar`, apparently because my python version is too new ?  I'm using the latest full Anaconda distro installation, which uses Python 3.7.4.  Any insight?

```
server@vav16:~$ conda install -c omnia pymbar
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: failed with repodata from current_repodata.json, will retry with next repodata source.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: | 
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed                                                                                                                                                                                      
                                                                                                                                                                                            
UnsatisfiableError: The following specifications were found                                                                                                                                 
to be incompatible with the existing python installation in your environment:

Specifications:

  - pymbar -> python[version='2.7.*|3.5.*|3.6.*']
  - pymbar -> python[version='>=2.7,<2.8.0a0|>=3.5,<3.6.0a0|>=3.6,<3.7.0a0']

Your python: python=3.7

If python is on the left-most side of the chain, that's the version you've asked for.
When python appears to the right, that indicates that the thing on the left is somehow
not available for the python version you are constrained to. Note that conda will not
change your python version to a different minor version unless you explicitly specify
that.

The following specifications were found to be incompatible with each other:



Package ca-certificates conflicts for:
python=3.7 -> openssl[version='>=1.1.1g,<1.1.2a'] -> ca-certificates
pymbar -> python=3.5 -> ca-certificates
Package wheel conflicts for:
pymbar -> python=3.5 -> pip -> wheel
python=3.7 -> pip -> wheel
Package pip conflicts for:
pymbar -> python=3.5 -> pip
python=3.7 -> pip
Package certifi conflicts for:
pymbar -> python=3.5 -> pip -> setuptools -> certifi[version='>=2016.09|>=2016.9.26']
python=3.7 -> pip -> setuptools -> certifi[version='>=2016.09|>=2016.9.26']
Package setuptools conflicts for:
python=3.7 -> pip -> setuptools
pymbar -> python=3.5 -> pip -> setuptools
```

vvoelz:
[...]

PR #404, opened 2020-07-19
change anaconda channel from omnia to conda-forge

chrisjonesBSU:
Addresses issue #396 

Per the comment by @Lnaden, conda-forge should be the channel used when installing pymbar via anaconda; however, the installation instructions in the getting started section still specify to use omnia.

I ran across this issue due to being unable to install via omnia when using python 3.7


jaimergp:
LGTM!

Issue #409, opened 2020-11-05
Update GHA


jaimergp:
Switch to `setup-miniconda` and backport to `master`.

PR #410, opened 2020-11-10
Update GHA to use Conda-Incubator

Lnaden:
Implements the Conda-Incubator to GitHub Actions instead of old
activate script. Removes the older helper scripts.

Partial resolution to #409


jaimergp:
LGTM.

I think we can also get rid of `devtools/{travis-ci,appveyor}`, as well as there corresponding yml files at the repo root.

Lnaden:
I don't want to drop the other tools yet until we are certain we've pulled what we need from them for the documentation compile and upload.

jaimergp:
Oops, I realized a bit too late that we need setup-miniconda **v2** 😬

.github/workflows/CI.yaml
```diff
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macOS-latest, ubuntu-latest, windows-latest]
-        python-version: [3.7, 3.8]
+        python-version: [3.8, 3.9]
```
jchodera, line 26: conda-forge is still building for 3.6 through 3.9---should we test with the same versions?

.github/workflows/CI.yaml
```diff
@@ -46,34 +46,33 @@ jobs:
...
 
-    - name: Create environment for package
-      shell: bash
-      run: |
-        . devtools/github-actions/initialize_conda.sh
-        conda activate
-        conda info
-        python devtools/scripts/create_conda_env.py -n=test -p=$PYVER devtools/conda-envs/test_env.yaml
+    - name: Setup Conda
+      uses: conda-incubator/setup-miniconda@v1
```
jchodera, line 50: The OpenFF folks were just talking about `v2` update here: https://github.com/openforcefield/status/issues/13

Lnaden: I bumped that in #411 since I missed it in this one.

.github/workflows/CI.yaml
Lnaden: Good question. I don't think we need to though. We want to make sure we are supporting the latest versions and following the protocol for "Most recent 2 Python Versions." I can understand supporting the "Most Recent 2"-1 version (3.7) in this case while everyone else gets pulled up. 

We can specify in the `meta.yaml` on conda-forge to restrict to Python >= 3.8 and it just wont build the lesser versions.

We also do not actually do anything which is Python 3.6 incompatible, I think, so there's no harm yet (not like we have any Walrus in the code yet).

jaimergp: Maybe we don't need the full matrix for all Python versions but can still include Ubuntu. Also there's a chance `pymbar` v4 goes `noarch` if the C extension is dropped, so conda-forge would be simplified.

Lnaden: I like having the full matrix in our own CI testing since it lets us see upstream dependency issues. But I agree, if we can go `noarch` then the conda-forge build will be much simpler.

PR #411, opened 2020-11-10
Bump conda-incubator in GHA to v2

Lnaden:
As it says on the tin. Follow up to #410


Lnaden:
So I cannot reproduce these failures locally in osx or ubuntu docker images. They are semi-stochastic on the GHA, but they always seem to fail at least one of them.

I've added some debug lines to at least print the conditions when it fails.

mrshirts:
Failures in test_logsumexp and test_logsumexp_b look like too fine a sensitivity to rounding errors.   Not sure if there is a way to adjust that with arguments to assert_equal - maybe will need to be adjusted to assert_almost_equal?

Lnaden:
>  maybe will need to be adjusted to assert_almost_equal?

That is what I was going to switch to, but I was more bothered by the fact I can't reproduce it locally, yet it so frequently fails on the GHA. I'm also trying to chase down the errant SVD error in the mbar solvers test.

Lnaden:
That seems to have worked. the Black failures are odd since I dont have them locally. I'm merging this to get it in.

PR #413, opened 2020-11-12
Modernize setup.py and versioning

### LINKED ISSUE #377: Revisit setup.py and consider versioneer
In case we forget about this!

jaimergp:
Changes:

* Added versioneer support
* Modernized setup.py

Questions:

- [x] Are we shipping the C extension in pymbar 4?

Closes #377


Lnaden:
I don't think we are building the C extension any more. I think the JIT work @jchodera has been working on is meant to replace it. In the existing version, it was only ever built in Python 2.X.

mrshirts:
Correct, no C extension.  I'm the one that is still a bit behind on the jit extension - I had it working in JAX, but not in numba yet. Will be working on that the next couple of days.

jaimergp:
The linting errors are for old_mbar.py, which should be deprecated soon (#415), so no need to worry. Merging!

jaimergp:
The version looks funny now, but we'll fix that with proper tags for v4.

Issue #364, opened 2020-02-25
List of what to backport to pymbar3 in preparation for pymbar4 release


mrshirts:
1. Warning about deprecation of old names so people are aware the change for pymbar4

2. Memory leak fixes (When addressed correctly in #359

mrshirts:
3. Bug fix in BAR caught in #350.

jaimergp:
I think we should also add a warning on the new `pmf.py` module and the `PMF` class, which now address the former `mbar.computePMF`?

mrshirts:
What specific warning?

jaimergp:
In `pymbar v3`, PMF is computed via `mbar.MBAR.computePMF()`. The deprecation warning in this case should redirect to `pmf.PMF(...)`, if I understood correctly?

mrshirts:
We can have such a message: "when you use pymbar4, use pmf.PMF functionality" though the syntax is not quite the same, and it can't be used in pymbar4, which is different from the messages on snake_case, since it's set up so that you can use new syntax immediately.

jaimergp:
> Memory leak fixes (When addressed correctly in #359)

359 got closed, but issue #341 is still open. Will we backport anything to v3 or this is not part of the roadmap?

mrshirts:
Nothing from #341 will be backported to 3.0.

jaimergp:
See #417 for a summary of changes. Closing this one for clarity!

PR #414, opened 2020-11-13
Backport PR 350

jaimergp:
Bring #350 and fixes to v3 -- addresses https://github.com/choderalab/pymbar/issues/364#issuecomment-592150921

I have also added deprecation warnings to the future behavior of `return_dict` here. Is that going to be a keyword option in v4 or just the default behavior with no flags?

I might have also fixed a bug in 3.0.5, where return_dict was still True by default in one method.

### DIFF
--- a/pymbar/bar.py
+++ b/pymbar/bar.py
@@ -46,6 +46,7 @@ __license__ = "MIT"
 #=============================================================================================
 # IMPORTS
 #=============================================================================================
+import warnings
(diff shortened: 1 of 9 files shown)


jaimergp:
@Lnaden @mrshirts ready for review

mikemhenry:
Now that we have pymbar4 out, this seems doesn't seem to be needed anymore

PR #418, opened 2020-11-17
Update README installation instructions from -c omnia to -c conda-forge.

bdice:
It looks like conda-forge is the intended method of installation for this package, based on the docs. This PR updates the README to use conda-forge instead of omnia. (I just happened to check for a conda-forge package before installing.)

Users installing from omnia would get an outdated version 3.0.3 instead of conda-forge's 3.0.5. I also updated the badge URLs. The conda-forge download count is much higher! 😄

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 [![Build Status](https://travis-ci.org/choderalab/pymbar.png)](https://travis-ci.org/choderalab/pymbar)
-[![Anaconda Cloud Downloads Badge](https://anaconda.org/omnia/pymbar/badges/downloads.svg)](https://anaconda.org/omnia/pymbar)
-[![Anaconda Cloud Badge](https://anaconda.org/omnia/pymbar/badges/installer/conda.svg)](https://anaconda.org/omnia/pymbar)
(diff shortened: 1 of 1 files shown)


jaimergp:
Thanks! 

We are releasing [pymbar4 ](https://github.com/choderalab/pymbar/tree/pymbar4) and [3.0.6](https://github.com/choderalab/pymbar/tree/3.0_branch) soon. I think those branches include these changes, so I'll leave it to the other maintainers whether we merge this or not!

We might copy the conda-forge packages to `omnia` in the meantime as a compromise to avoid the misleading versions.

Thoughts @Lnaden @mrshirts @jchodera ?

Lnaden:
> We might copy the conda-forge packages to omnia in the meantime

I would not. We want to steer people away from the omnia channel if we can, and backporting I feel will encourage that. `pymbar` has been one of the first packages we ported off of Omnia way back, lets keep it off there. The old versions are still on Omnia for reproducibility, but will slowly stop being fetched on accident since we stopped building them for newer python versions some time ago.

mikemhenry:
Any holdup on merging this PR?

Lnaden:
There shouldn't be. I'll merge it.

bdice:
Thank you @mikemhenry @Lnaden!

Issue #419, opened 2020-11-25
Should have \sum_n W_nk = 1 error with pymbar3.0.5 while analysis is working with pymbar3.0.3


hannahbaumann:
Hi,

I'm analyzing free energy calculations (Gromacs-based) using alchemlyb. The analysis fails due to following error:
Warning: Should have \sum_n W_nk = 1.  Actual column sum for state 0 was 0.835621. 21 other columns have similar problems.
The pymbar version here was 3.0.5

Running the same input files/script in a different environment with pymbar==3.0.3 didn't raise this error, the script was able to calculate the free energy difference. 

The issue seems to be similar to other issues, see e.g. https://github.com/choderalab/pymbar/issues/320

Here is a folder with the .xvg files from one of the runs and a python script for the analysis and the dependencies in the 2 different environments: 
https://drive.google.com/drive/folders/1lKncHmRj38OY8H2YjsnYHdIpEcvxKKpX?usp=sharing
Let me know if you need more information!

davidlmobley:
Tagging @mrshirts

jaimergp:
This is the diff between two versions: https://github.com/choderalab/pymbar/compare/3.0.3...3.0.5

This [particular change](https://github.com/choderalab/pymbar/compare/3.0.3...3.0.5#diff-f5541a3c868c7773917580b717127325539785ff68b367eebe1427e822c878e6R383) in the default tolerance might be interesting to analyze.

Otherwise, the standard approach here is to run `git bisect run ...` between the two tags and let the script exit code reveal the offending commit. For this to happen you'll need a local copy of the repo, and fresh environment where `pymbar` is installed in _editable_ mode. Roughly:

```
# get pymbar env file for the 3.0.x branch
% wget https://github.com/choderalab/pymbar/raw/3.0_branch/devtools/conda-envs/test_env.yaml
% conda env create -n pymbar-dev test_env.yaml
% conda activate pymbar-dev
% git clone https://github.com/choderalab/pymbar
% cd pymbar
% git checkout 3.0.3  # start with the working state
% pip install -e .
% git bisect start 3.0.3
% git bisect run your_script.py
```

xiki-tempula:
This is similar to the issue that I got in #407

jaimergp:
Ok, I think I have found the issue.

* ✅ Last commit where this was working is fe9ea5323c5b2deafe25879a6dfb334822f6c1be
[...]

Issue #320, opened 2019-12-04
Should have \sum_n W_nk = 1


bergazin:
Hello,

I'm trying to extract the free energy from a number of yank experiments and some of them are throwing this error:

```
Traceback (most recent call last):
  File "/export/home/bergazin/anaconda3/envs/sampl-2/bin/yank", line 11, in <module>
    load_entry_point('yank==0.23.7', 'console_scripts', 'yank')()
  File "/export/home/bergazin/anaconda3/envs/sampl-2/lib/python3.7/site-packages/yank/cli.py", line 73, in main
    dispatched = getattr(commands, command).dispatch(command_args)
  File "/export/home/bergazin/anaconda3/envs/sampl-2/lib/python3.7/site-packages/yank/commands/analyze.py", line 147, in dispatch
    single_run()
  File "/export/home/bergazin/anaconda3/envs/sampl-2/lib/python3.7/site-packages/yank/mpi.py", line 271, in _wrapper
    return run_single_node(rank, task, *args, **kwargs)
  File "/export/home/bergazin/anaconda3/envs/sampl-2/lib/python3.7/site-packages/yank/mpi.py", line 220, in run_single_node
    result = task(*args, **kwargs)
  File "/export/home/bergazin/anaconda3/envs/sampl-2/lib/python3.7/site-packages/yank/commands/analyze.py", line 142, in single_run
    output = analyze.analyze_directory(args['--store'], **analyzer_kwargs)
  File "/export/home/bergazin/anaconda3/envs/sampl-2/lib/python3.7/site-packages/yank/analyze.py", line 768, in analyze_directory
    analysis_data = auto_experiment_analyzer.auto_analyze()
  File "/export/home/bergazin/anaconda3/envs/sampl-2/lib/python3.7/site-packages/yank/analyze.py", line 172, in make_copy
    return copy.deepcopy(wrap(*args, **kwargs))
  File "/export/home/bergazin/anaconda3/envs/sampl-2/lib/python3.7/site-packages/yank/analyze.py", line 681, in auto_analyze
    _ = self.get_experiment_free_energy_data()
  File "/export/home/bergazin/anaconda3/envs/sampl-2/lib/python3.7/site-packages/yank/analyze.py", line 172, in make_copy
    return copy.deepcopy(wrap(*args, **kwargs))
  File "/export/home/bergazin/anaconda3/envs/sampl-2/lib/python3.7/site-packages/yank/analyze.py", line 541, in get_experiment_free_energy_data
[...]

PR #425, opened 2021-08-05
Forward maximum_iterations from MBAR constructor

maxentile:
Possibly related to https://github.com/choderalab/pymbar/issues/320 , https://github.com/choderalab/pymbar/issues/419 . In these issue reports, the logs suggest that the default `maximum_iterations = 250` was reached.

However, the default `maximum_iterations` in the MBAR constructor https://github.com/choderalab/pymbar/blob/3c4262c490261110a7595eec37df3e2b8caeab37/pymbar/mbar.py#L73 is much larger than 250.

The `maximum_iterations` argument in the MBAR constructor is currently unused. When `maximum_iterations` is absent from the `options` dictionary that is forwarded from the MBAR constructor to the `adaptive` solver, `maximum_iterations` defaults to 250 here https://github.com/choderalab/pymbar/blob/3c4262c490261110a7595eec37df3e2b8caeab37/pymbar/mbar_solvers.py#L271 

This patch solves the minimal example provided in this comment https://github.com/choderalab/pymbar/issues/419#issuecomment-796230485 (which worked in 3.0.3 but failed in 3.0.5), although other examples in the thread may not be addressed by this patch.

### DIFF
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -296,6 +296,7 @@ class MBAR:
         for solver in solver_protocol:
             if 'options' not in solver:
                 solver['options'] = dict()
+                solver['options']['maximum_iterations'] = maximum_iterations
             if 'verbose' not in solver['options']:
                 # should add in other ways to get information out of the scipy solvers, not just adaptive,
                 # which might involve passing in different combinations of options, and passing out other strings.
--- a/pymbar/mbar_solvers.py
+++ b/pymbar/mbar_solvers.py
@@ -268,7 +268,7 @@ def adaptive(u_kn, N_k, f_k, tol = 1.0e-12, options = None):
     """
     # put the defaults here in case we get passed an 'options' dictionary that is only partial
     options.setdefault('verbose',False)
-    options.setdefault('maximum_iterations',250)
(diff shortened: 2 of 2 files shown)


jchodera:
Thanks for the fix, @maxentile!

I'm not familiar with the pymbar 3.* codebase, so tagging @mrshirts and @Lnaden for review.

Lnaden (approved):
Good catch and debugging, @maxentile. I would like to make another request. As @mrshirts pointed out in #419, the max iterations in the default is too low in general, can we increase that up to some higher value as well, maybe even the default from the `MBAR` constructor of 10000? Would that cause issues?

maxentile:
>As @mrshirts pointed out in #419, the max iterations in the default is too low in general, can we increase that up to some higher value as well, maybe even the default from the MBAR constructor of 10000? Would that cause issues?

Not sure if there's a reason to have the default `maximum_iterations` in the `MBAR` constructor differ from the default `maximum_iterations` in the `adaptive` solver. Assuming not, the requested change is applied in https://github.com/choderalab/pymbar/pull/425/commits/bbadb5b4608de3f139a1070e33270d63e697d1fa .

Issue #429, opened 2022-03-01
Add tests for BAR methods


jchodera:
We appear to have completely neglected to test any of the BAR functionality in `pymbar/bar.py`.

PR #430, opened 2022-03-01
Fix #427 : Add BAR compute overlap method

### LINKED ISSUE #427: Add methods to compute overlap and effective sample sizes for BAR?
@mrshirts : What do you think about adding a corresponding `computeOverlap()` method for BAR? This could help users identify when the BAR uncertainty estimate will be unreliable.

jchodera:
This PR also adds minimal tests for BAR, which appear to have been accidentally omitted (#429).

cc @zhang-ivy

### DIFF
--- a/pymbar/__init__.py
+++ b/pymbar/__init__.py
@@ -30,7 +30,7 @@ __email__ = "levi.naden@choderalab.org,michael.shirts@colorado.edu,john.chodera@
 
 from pymbar import timeseries, testsystems, confidenceintervals, version
 from pymbar.mbar import MBAR
-from pymbar.bar import BAR, BARzero
+from pymbar.bar import BAR, BARoverlap, BARzero
 from pymbar.exp import EXP, EXPGauss
 import pymbar.old_mbar
 
--- a/pymbar/bar.py
+++ b/pymbar/bar.py
@@ -147,7 +147,7 @@ def BARzero(w_F, w_R, DeltaF):
     return fzero
 
 
-def BAR(w_F, w_R, DeltaF=0.0, compute_uncertainty=True, uncertainty_method='BAR',maximum_iterations=500, relative_tolerance=1.0e-12, verbose=False, method='false-position', iterated_solution=True, return_dict=False):
+def BAR(w_F, w_R, DeltaF=0.0, compute_uncertainty=True, uncertainty_method='BAR', maximum_iterations=500, relative_tolerance=1.0e-12, verbose=False, method='false-position', iterated_solution=True, return_dict=False):
     """Compute free energy difference using the Bennett acceptance ratio (BAR) method.
 
     Parameters
@@ -174,7 +174,7 @@ def BAR(w_F, w_R, DeltaF=0.0, compute_uncertainty=True, uncertainty_method='BAR'
     method : str, optional, defualt='false-position'
         choice of method to solve BAR nonlinear equations, one of 'self-consistent-iteration' or 'false-position' (default: 'false-position')
     iterated_solution : bool, optional, default=True
-        whether to fully solve the optimized BAR equation to consistency, or to stop after one step, to be 
+        whether to fully solve the optimized BAR equation to consistency, or to stop after one step, to be
         equivalent to transition matrix sampling.
     return_dict : bool, default False
         If true, returns are a dict, else they are a tuple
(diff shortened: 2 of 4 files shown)


mrshirts:
Thanks for adding the tests!

I wonder if it would be best to not have to create an MBAR instance?  I believe the inside of the ( )^-1 in equation 10b in Bennett's 1976 paper is equivalent to the MBAR, and could be perhaps used instead of the overhead of MBAR?  Thoughts?

jchodera:
> I wonder if it would be best to not have to create an MBAR instance? I believe the inside of the ( )^-1 in equation 10b in Bennett's 1976 paper is equivalent to the MBAR, and could be perhaps used instead of the overhead of MBAR? Thoughts?

My thinking was that establishing the (experimental) feature and tests is probably fine for now, and we can optimize for performance later.

I should probably add a warning that the method is experimental for now and the API may change in future. Do we have a standard warning for that already?

jchodera:
@mrshirts : Can you review this so we can merge it and get a new version cut for @nividic and @cbayly13?

@mikemhenry can help cut a new release and update the conda-forge recipe.

mrshirts:
Just FYI, I am looking at this, and working to trace down a small discrepancy in different definitions of overlap.

mrshirts:
OK, I think this can go in.   A couple of thoughts:

* We probably want to think about making the BAR and MBAR interfaces more consistent.  That is later.  I'll file an issue here.
* There are some really interesting near-equalities in definitions of the overlap matrix. This is explored a bit in the section "The overlap matrix is a quantitative estimator of the phase-space overlap" in https://link.springer.com/article/10.1007/s10822-015-9840-9, which gives an relationship between "BAR" and "MBAR" ways of thinking about things which ends up not quite being an equality, but are different by a factor proportional to 1/N. The paper doesn't quite explore things enough. It appears to be related to https://github.com/choderalab/pymbar/issues/281, as the overlap integral is a key part of the variance estimates (in a complicated way, so not attempting to give the result here). 

SO, If one wants to ask the question, "Is the overlap good or bad"? then either method works great.  So if the need is to have this feature, then go ahead and merge it in.  Some research to do longer term on what the "right" way is.  Probably the one that is consistent with the uncertainty estimate.

mrshirts (approved):
Looks good, see comments above about what BAR overlap actually means.

jchodera:
> We probably want to think about making the BAR and MBAR interfaces more consistent. That is later. I'll file an issue here.

This might be useful, but if we really wanted consistent interfaces, the MBAR would be sufficient to cover all circumstances.

> SO, If one wants to ask the question, "Is the overlap good or bad"? then either method works great. So if the need is to have this feature, then go ahead and merge it in. Some research to do longer term on what the "right" way is. Probably the one that is consistent with the uncertainty estimate.

We do need something like this, so I'll go ahead and merge this.

Perhaps we could later add an optional `method` argument that could select the exact method used for overlap computation.

jchodera:
@mrshirts : Would it help for this method to warn that this API is experimental before we release?

mrshirts:
Clarify "this API". The whole MBAR API? Just BARComputeOverlap?

jchodera:
> Clarify "this API". The whole MBAR API? Just BARComputeOverlap?

I just meant that if we wanted to unify the way in which the class is initialized, there would be no need to have a separate BAR class---the MBAR class would provide a superset of the functionality of BAR.

mrshirts:
I think that sometimes people would have w_F and w_R, and having a BAR interface would be useful.  It could potentially be MBAR under the hood, though on the other hand IIRC the BAR solver is more numerically stable.

jchodera:
> I think that sometimes people would have w_F and w_R, and having a BAR interface would be useful. It could potentially be MBAR under the hood, though on the other hand IIRC the BAR solver is more numerically stable.

@mrshirts : Let's open a new issue to discuss those API extensions.

@mikemhenry : Can you add a warning to the `BARoverlap` method like "Warning: This API is experimental and subject to change in future releases"? Then we can cut a release!

nividic:
@jchodera Thank you so much to get these changes in. @mikemhenry we are really hoping to incorporates this in our analyses asap. We do not mind if the API will change in the future; it is an inconvenience that we can address later. Please let us know when a release is cut with this feature in

mikemhenry:
@nividic will do! In the meantime, you can install from source to get these changes now with `python -m pip install git+https://github.com/choderalab/pymbar.git`

PR #439, opened 2022-06-17
With 0logfactor

mrshirts:
This will need to be rewritten, but posting it here to show what should be added.  See the commits on Sept 29,2020 for the bug fix.


mrshirts:
Abandoned in favor of #441

PR #441, opened 2022-06-17
Fix for expectations of very small numbers

mrshirts:
Previously, expectations of very small numbers would not be correct.  Before, the code ensured that logarithms of expectations were greater than 1 by shifting the values.  Now, we only make sure they are greater than zero. this does not appear to have any numerical issues (though it's possible there could be underflow), and fixes the previous problems.  Leaving it in as factor instead of hardcoding in case we need to make it bigger in the future).

### DIFF
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -645,6 +645,11 @@ class MBAR:
 
         """
 
+        logfactor = 0  # make sure all results are larger than this number.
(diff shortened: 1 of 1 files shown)


mrshirts:
We probably don't want the user to change it.  I think 0 is the right value, but maybe something really small but greater than zero is better in some cases?  So I was just leaving it as a variable for now.

mrshirts:
Merging this in.

PR #442, opened 2022-06-20
[WIP] Trying to make MBAR work better adaptively when it fails.

mrshirts:
* Changed the way defaults are set up.  Now the default is adaptive + LBFGSB. 
* Previously, it would run through solvers iteratively, assuming that multiple refinement is needed.  However, if a solver is good, it should solve in one go.  Therefore, we now loop through solvers until we find one that works. 
* So now, automatically tries both adaptive and L-BFGS-B if the first one chosen fails.
* If multiple solvers are tried and fails, returns the best result.
* Also makes the outputs and options of 'adaptive' more consistent with scipy methods to make the logic simpler.

TODO: 
* scipy minimizer doesn't appear to be recognizing tolerance correctly - it will state success when the norm of the gradient is still relatively large (or at least larger than the tolerance)
* Verify this
* Decide the order of the methods in default - should -LBFGS-B be tried first? 
* Maybe it's better to actually run multiple methods; start with something fast, then conclude with adaptive, which we know is more robust, once it's relatively close. 
* Make the output good enough to use/understand for now (we will move to logging system in 4.0). 
* See how different solvers behave and get tolerance to work correctly with them.

### DIFF
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -1,9 +1,9 @@
 ##############################################################################
 # pymbar: A Python Library for MBAR
 #
-# Copyright 2017 University of Colorado Boulder
-# Copyright 2010-2017 Memorial Sloan-Kettering Cancer Center
-# Portions of this software are Copyright (c) 2010-2016 University of Virginia
+# Copyright 2016-2022 University of Colorado Boulder
+# Copyright 2010-2022 Memorial Sloan-Kettering Cancer Center
+# Portions of this software are Copyright (c) 2010-2015 University of Virginia
 # Portions of this software are Copyright (c) 2006-2007 The Regents of the University of California.  All Rights Reserved.
 # Portions of this software are Copyright (c) 2007-2008 Stanford University and Columbia University.
 #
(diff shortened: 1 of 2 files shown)


orbeckst:
You're probably aware of @xiki-tempula 's implementation of  "[AutoMBAR](https://alchemlyb.readthedocs.io/en/latest/estimators/alchemlyb.estimators.MBAR.html#alchemlyb.estimators.AutoMBAR)" in alchemlyb. He did the testing and ultimately came up with the rationale for the order of methods that are tried.

mrshirts:
Yes, I looked at that, but wasn't sure about logic.  "hybr" is interesting since it is a root-finding method, which  we haven't really looked at before.  

Each method seems to handle tolerances a bit differently, and that's one thing that I've been trying to get to the bottom of.  Calling adaptive at the end, which tends to continue a little further than other methods, seems to be a good way to makes sure that the results end up accurate, but I don't know if that's the "best" way to go.

xiki-tempula:
@mrshirts I called `hybr` first as it is easy to fail but it seems to be super fast. So calling it first to save time on the easy cases.
If `hybr` failed, I will call `adaptive` which could solve 99% of the cases.
Lastly, I will call `BFGS`. It is not as robust as `adaptive` but could solve the cases that `adaptive` cannot.

PR #444, opened 2022-06-27
Syncing Pymbar4 with master

mrshirts:
There are enough changes between pymbar4 and master that I had to manage things manually.  

There were also cases in the test suite where logfactor=0 resulted in some divide by zero factors, so I made logfactor = 4 x eps, where eps = np.finfo(np.float64).eps, which appears to have resolved almost all errors.  This MAY cause problems for expectations that are lower than 4 x eps, will need to be studied a bit more.  Generally, this will just apply to rare occupancies.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -129,9 +129,11 @@ jobs:
       run: |
         pylint $PACKAGE/
 
-    # TODO: Remove --exclude pymbar/old_mbar.py after deprecation
+    # TODO: Remove old_mbar.py from --extend-exclude after deprecation
+    # Black has a default --exclude which reads .gitignore.
+    # Black's --exclude and --extend-exclude are single specified RegEx patterns and overwrite if multi-marked.
     - name: Run black check
       shell: bash -l {0}
       if: always()
       run: |
-        black --check -l 99 $PACKAGE/ examples/ --exclude pymbar/old_mbar.py --exclude $PACKAGE/_version.py
+        black --check -l 99 $PACKAGE/ examples/ --extend-exclude "$PACKAGE/(old_mbar.py|_version.py)"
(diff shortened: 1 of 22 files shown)


mrshirts:
Addressed some lint issues.  There may be some version mismatch issues with black, may address those later.

mrshirts:
Looks like everything passes except lint (and I pre-linted some of these files already, so need to match up versions), so that's probably OK for now.  Will address lint issues before final v4 release.

mrshirts:
@Lnaden, any thoughts on the linting error?

Lnaden:
The `pylint` bits I should have fixed with the first commit here. They were mostly complaining about a `def` inside of a loop using a variable from that loop. 

The `black` errors appear to be mostly caused by `var ** exponent` where its complaining about the spaces around the `**` operator for things not surrounded by `( )`. I think I can fix those real quick

Lnaden:
This should fix all the lint problems. Fun fact I just learned: `black`'s `--exclude` has a default value of what's in the `.gitignore` file, and its not multi-mark! So `--exclude fileA -- exclude fileB` only reads the `fileB` as a RegEx pattern, overwriting the default and the first one. They have `--extend-exclude` and its also not multi-mark, so I had to make a new regex pattern to match the 2 files we are trying to ignore. I also added these comments to the CI file for the next person.

mrshirts:
Looks like fes.py is not working still.  When I run it through black, it does not change it. 

Hmm.  what version of black are you using?  Looks like I'm using version 19.10b0?  Where do I find the version that is being used?

mrshirts:
Eh, no, I upgraded black to 22.6 and it still doesn't change fes.py

Lnaden:
22.6.0, you can see it in the earlier action step: https://github.com/choderalab/pymbar/runs/7267054323?check_suite_focus=true#step:4:25

I found the typo I had. I had the wrong window in focus and typed something in, then overzealously pressed ctrl+z too many times. This should do it.

mrshirts:
OK, I'm going to go ahead and merge since there are no objections.

PR #446, opened 2022-06-29
Pymbar 4.0 adaptive logic

mrshirts:
Pass at adding adaptive logic to better converge mbar estimates.  Now passes pytest test.  Could use a bit more cleaning up, but basic logic is working.


mrshirts:
OK, this appears to have failed with functions returning the error messages that aren't expected.  I will check this . . . Interesting that it passed when running pytest locally, but not here.

devtools/conda-recipe/README.md
```diff
@@ -6,7 +6,7 @@ it, running the tests, and then if successful pushing the package to binstar
 (and the docs to AWS S3). The binstar auth token is an encrypted environment
 variable generated using:
 
-binstar auth -n repex-travis -o omnia --max-age 22896000 -c --scopes api:write
+binstar auth -n repex-travis -o conda-forge --max-age 22896000 -c --scopes api:write
```
jchodera, line 9: I don't think we're using `binstar` anymore. 

Does this repo need to be updated to the current [MolSSI cookiecutter?](https://github.com/MolSSI/cookiecutter-cms)

examples/constant-force-optical-trap/force-bias-optical-trap.py
```diff
@@ -224,7 +224,7 @@ def main():
     # Initialize MBAR.
     print("Running MBAR...")
     # TODO: change to u_kn inputs
-    mbar = pymbar.MBAR(u_kln, N_k, verbose=True, relative_tolerance=1.0e-10)
+    mbar = pymbar.MBAR(u_kln, N_k, verbose=True, relative_tolerance=1.0e-10, solver_protocol = 'robust')
```
jchodera, line 227: Shouldn't `robust` be default?

examples/constant-force-optical-trap/output/counts-expected-0.out
```diff
@@ -1,50 +1,50 @@
- 494.896      3701.003313        58.541052
```
jchodera, line 1: What's the point of including "expected output" targets if this changes even if we are only changing the efficiency and robustness, rather than the expected result? Shouldn't we just adjust the comparison tolerance and keep a single, well-converged set of results here?

pymbar/mbar_solvers.py
```diff
@@ -3,18 +3,49 @@
...
+    dict(method="BFGS", continuation=True),
+    dict(method="adaptive", options=dict(min_sc_iter=0)),
+)
+
+DEFAULT_SOLVER_PROTOCOL = (
+    dict(method="hybr", continuation=True),
+    dict(method="adaptive", options=dict(min_sc_iter=0)),
+)
+
+ROBUST_SOLVER_PROTOCOL = (
```
jchodera, line 23: Shouldn't "robust" be default, with faster methods available if desired?

pymbar/mbar_solvers.py
```diff
@@ -253,12 +284,11 @@ def adaptive(u_kn, N_k, f_k, tol=1.0e-12, options=None):
 
     options: dictionary of options
         gamma (float between 0 and 1) - incrementor for NR iterations (default 1.0).  Usually not changed now, since adaptively switch.
-        maximum_iterations (int) - maximum number of Newton-Raphson iterations (default 250: either NR converges or doesn't, pretty quickly)
+        maxiter (int) - maximum number of Newton-Raphson iterations (default 10000: either NR converges or doesn't, pretty quickly)
```
jchodera, line 287: Why change the name from something clear and comprehensible (`maximum_iterations`) to something shorter (`maxiter`)? Why change the name at all?

jchodera (commented):
Shouldn't the `robust` approach become the default? Other, faster approaches could be selected by the user.

examples/constant-force-optical-trap/output/counts-expected-0.out
mrshirts: Reproducibility, basically.  They should get the same answer as in the code.  

I don't know how to have git compare "up to a number".

pymbar/mbar_solvers.py
mrshirts: Good question.  `default` is essentially what was done before. I can play around with this a little.

examples/constant-force-optical-trap/force-bias-optical-trap.py
mrshirts: That's a good question.  I think this is to get the framework right for now, and the precise solvers used will need a bit more experimentation - what is "robust", and what is "default".  I'm totally find with using what we think is the most robust for the default, and then having a "faster" option.  I will set those solvers in a final PR.

devtools/conda-recipe/README.md
mrshirts: @Lnaden any thoughts on binstar or updating the cookiecutter?

PR #447, opened 2022-06-29
Pymbar4 with jax (real try now)

mrshirts:
A branch that supports JAX, including minimize using BFGS and adaptive. It does not pass all the tests yet (because some pathways are mixed JAX/numpy, and those don't get handled well). But it will pass a lot of of the tests locally.  I have not tested in GPU's yet.  But it at least can be configured to run with JAX now on the conda environment.  solver_protocol="jax" will exercise all of the JAX accelerated options.


Lnaden:
It should also be possible to just add `jax` to the dependency list to get it to start testing here with the CI. It'll at least work with CPU `jax` which is on `conda-forge`.

mrshirts:
> About this branch, is the plan to require JAX, or make it optional? If its optional, I can spend time to get this in a state where we have fewer codepaths by sharing namespaces.

**That's** what I don't know about.  alchemlyb developers would prefer not to have JAX.  But I don't know how to have conda install such that jax is optional.  Do we need to have multiple conda installs? Can you pass options?  (like having an option to turn jax install off?) 

In terms of the numpy functions, I think almost all of them have equivalents in both numpy and JAX (i.e. exp, etc.) So that can be done fairly easily.  The thing I don't know how to do yet: eliminate having number of jax_ jit_ versions of calls.

Lnaden:
>  But I don't know how to have conda install such that jax is optional. Do we need to have multiple conda installs? Can you pass options? (like having an option to turn jax install off?)

Can you have multiple package outputs from one recipe? Yes
Can you have optional dependencies? Not really (its very hacky and uses fields that are not meant to do it).

In either case, we can reduce the number of code paths with a `try... except...` at the start to get all of the namespaces to line up so the code all looks the same.

A couple different paths:
* Create a multi-output recipe which has a `pymbar-nojax` and a `pymbar` which depends on `pymbar-nojax` that is just `pymbar-nojax` + `jax` dependency. Pros: All in one place, but makes a separate pymbar package in the namespace. Con: anyone who wants to use it without jax needs `pymbar-nojax`; names subject to whatever we want
* Maintain different packages. Pro: Cleaner maintainer readability. Con: package in different space. 

> The thing I don't know how to do yet: eliminate having number of jax_ jit_ versions of calls.

We can decorate all the functions with a `functools.partial` [ala](https://jax.readthedocs.io/en/latest/jax-101/02-jitting.html#why-can-t-we-just-jit-everything)

```python
from functools import partial

@partial(jax.jit, static_argnames=['n'])
def g_jit_decorated(x, n):
  i = 0
  while i < n:
    i += 1
  return x + i

print(g_jit_decorated(10, 20))
```

but instead of just `@partial(jax.jit...)` we do `@partial(jit_if_available, ...)` which calls a function that checks for jit and then passes to it before returning func... or something.

Lnaden:
I'll work on making PR's to this PR in bite sized chunks to get the behavior you described there. List of what my PR's will do below:

* [ ] Make Decorator for JIT functions + Add JAX as dependency for GHA CI
* [ ] Overlap namespaces so all of the function calls are the same + check for JAX on import with `try...except`
* [ ] Align functions for with and w/o jax, modify decorator to check for JAX and JIT if so.
* [ ] Add global JAX disable option + write tests to compare JAX + not JAX outputs.

pymbar/mbar_solvers.py
```diff
@@ -145,16 +268,27 @@ def mbar_objective_and_gradient(u_kn, N_k, f_k):
...
+    logW = f_k - u_kn.T - log_denominator_n[:, newaxis]
+    W = exp(logW)
 
+    H = dot(W.T,W)
+    H *= N_k
+    H *= N_k[:, newaxis]
+    H -= diag(W.sum(0) * N_k)
+    return -1.0 * H
+
+jit_mbar_hessian = jax.jit(jax_mbar_hessian)
```
proteneer, line 291: Generally, best practice is to delay `jit` as late as possible.

pymbar/mbar_solvers.py
```diff
@@ -301,51 +447,58 @@ def adaptive(u_kn, N_k, f_k, tol=1.0e-12, options=None):
     # to calculate the first time.
     g = mbar_gradient(u_kn, N_k, f_k)  # Objective function gradient.
 
-    for iteration in range(0, options["maximum_iterations"]):
+    maxiter = options["maxiter"]
+    min_sc_iter = options["min_sc_iter"]
+    warn = "Did not converge."
+    for iteration in range(0, maxiter):
```
proteneer, line 453: you can actually `jit` a loop using `jax.lax.while_loop`:

https://jax.readthedocs.io/en/latest/_autosummary/jax.lax.while_loop.html`

proteneer: (if the number of iterations is known before hand, `jax.lax.fori_loop` is another option, but probably doesn't apply hear since we break adaptively)

pymbar/mbar_solvers.py
```diff
@@ -355,39 +508,77 @@ def adaptive(u_kn, N_k, f_k, tol=1.0e-12, options=None):
...
 
+    results = dict()
+    results["success"] = success
+    results["message"] = warn
+    results["x"] = f_k
+
+    return results
+
+
+def jax_core_adaptive(u_kn, N_k, f_k, gamma):
```
proteneer, line 543: you can probably wrap this entire function in a @jax.jit decorator instead

pymbar/other_estimators.py
```diff
@@ -531,6 +531,42 @@ def bar(
         return result_vals
 
 
+def bar_overlap(w_F, w_R):
+    """Compute overlap between foward and backward ensembles (using MBAR definition of overlap)
```
proteneer, line 535: typo foward

pymbar/mbar_solvers.py
```diff
@@ -1,20 +1,64 @@
...
 logger = logging.getLogger(__name__)
 
 # Below are the recommended default protocols (ordered sequence of minimization algorithms / NLE solvers) for solving the MBAR equations.
 # Note: we use tuples instead of lists to avoid accidental mutability.
-# DEFAULT_SUBSAMPLING_PROTOCOL = (dict(method="L-BFGS-B"), )  # First use BFGS on subsampled data.
-# DEFAULT_SOLVER_PROTOCOL = (dict(method="hybr"), )  # Then do fmin hybrid on full dataset.
-# Use Adpative solver as first attempt
-DEFAULT_SOLVER_METHOD = "adaptive"
-DEFAULT_SOLVER_PROTOCOL = (dict(method=DEFAULT_SOLVER_METHOD),)
+JAX_SOLVER_PROTOCOL = (
```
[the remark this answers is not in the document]
mrshirts: BFGS is currently the only one that jax.scipy.optimize.minimize supports.  We were trying to minimize the number of packages downloaded, and https://jaxopt.github.io/stable/index.html doesn't appear to have a conda installation (for people wanting to use it automatically, in validation, etc.).  BFGS combined with adaptive seems to work pretty well (on test cases so far - haven't been thorough yet!)

pymbar/mbar_solvers.py
proteneer: If you're using `jax.scipy.optimize` the entire optimization routine itself can probably be `jitted`. Generally you want to`jit` only once, and only at the outer most level.

Lnaden:
@proteneer Question about `jit` capability as an optional decorator you might know. We want to make JAX optional, which means `jit` has to be optional as well. I'd like to do this without duplicate code and without many `if` statements checking for `jax` existence.

Is it possible to wrap the `jit` decorator inside a separate decorator which calls something to the effect of:

```python
def jit_if_jax():
    def jitify(func):
        if JAX_FOUND:
            return jax.jit(func)
       return func
    return jitify

@jit_if_jax
def self_consistent_update(...)
    ...
```

That way the code inside `self_consistent_update` is identical if JAX is installed or not? Or does this break something in the way `jit` preps the code or re-jit's it every time the function is called?

proteneer:
it shouldn't break anything from `jax`'s point of view - as long as functions are traceable (via `jnp` ops as opposed to `np` ops) it should still be fully `jit`able

Lnaden:
> just fake a jit decorator in the non-jit

That's such a simpler solution with fewer pitfalls. Thank you.

Lnaden:
i have the following implemented on my own fork 

>  Overlap namespaces so all of the function calls are the same + check for JAX on import with try...except
 Align functions for with and w/o jax, modify decorator to check for JAX and JIT if so.

I will make a PR to this PR once #451 is approved and merged in here.

pymbar/mbar_solvers.py
mrshirts: I guess the issue is that different branches need to be jitted at different levels.  

I will try to at least jit around the loop in adaptive, or jit the entire function, though!

mrshirts:
> CI here fails on windows because there is no pre-packaged jax for windows. I'll figure out how to skip those in CI.

Probably illustrates a good reason to make jax optional.

Lnaden:
The branches got changed again, which complicates things a bit. I don't know what the desired logic is in choosing the solver in `mbar.py` between default, robust, jax, and bootstrap. Can you take a look and/or explained desired outcome, @mrshirts?

mrshirts:
Yeah, sorry for the issues.  Is the best thing for me to do to adjust things by resolving this merge conflict?  I can try to do that now.  

Based on that "first, don't fail" approach, I think we probably want "default" (which may eventually get set to robust). That will still be accelerated by Jax (because gradient, etc. are jitted), but not as much as JAX?   

Then we can provide instructions for speeding it up in the documentation (see #455)

mrshirts:
Had to reconcile my pymbar4_new_jax first. . . now working on the merge conflict.

mrshirts:
Whoops, didn't fix all tests. One sec.

Lnaden:
If you're happy with the logic, I can work on the JAX/syntax issues.

mrshirts:
Yeah, I'm good with the logic for now.  Have to head to a poster session, but will be back in a bit to check if there are other questions!  

the non _jax error looks like it's a stochastic issue with the example having histogram with zero population.  Maybe need to be fixed by decreasing the spacing.  I'll come back to it before release if 4.0 if it keeps happening.

Lnaden:
At a functional scientific level, this is ready, but I am cleaning up some test speeds since I doubled the number of tests, and I still need to reconfigure the dependencies in setup.py to require jax, and the tests to remove jax where appropriate.

mrshirts:
Cool, let me know when I should start verifying things scientifically.

Lnaden:
@mrshirts this should be ready to go. I've got all my changes in, including adding some additional samples int he 2d FE surface of examples to reduce chance of empty bins. (300/oscillator -> 500/oscillator)

pymbar/other_estimators.py
mrshirts: Thanks!

mrshirts:
It appears the failures are linting (where I ran black locally, and no complaints, so I still don't know what is going on), and 'trust-exact' failing with nans, which is weird since it only happens in Windows.   So ALMOST ready to merge.

Lnaden:
I marked that one test which was failing flaky since it appeared to be stochastic on windows only (I could not reproduce and when I did a black+push the code passed on GitHub Actions as well). That one parameter is allowed to fail once before raising an actual error.

Linting should be fixed as well.

If that was the last outstanding issues, and if the tests pass, I think that may have been the last items.

PR #448, opened 2022-07-05
Pymbar4 bootstrapping

mrshirts:
Adding support for bootstrapped errors to all expectations.

Computes bootstrap uncertainties and covariances for:

- compute_free_energy_differences
- compute_expectations
- compute_multiple_expectations
- compute_perturbed_free_energies
- compute_entropy_and_enthalpy

These are accessed by using the uncertainty_method = "bootstrap". Note that MBAR needs to be initialized with "n_bootstrap=(some integer)" samples in order to do use this method - an error is omitted otherwise.  Possibly could be done differently, but the logic is simpler this way for now. 

I could use a little help in rewriting the pytests.  Currently, to get bootstrap uncertainties from various functions, as one will now want to check both the default uncertainties and the bootstrap uncertainties, and MBAR needs to be initialized differently each way, which the pytests don't seem to handle.

Note this is built off of #444, so many of the differences are from that branch.


zhang-ivy:
@mrshirts : I should be able to test this out in the next couple days. 

What's your expected timeline for getting this merged and pymbar 4 released? The reason I ask is because I am generating data for a paper and ideally I would be able to use a stable version of pymbar to do the bootstrapping. 

@mikemhenry is planning on fixing the pymbar 3 unit tests (which will take about a day) so that we can get https://github.com/choderalab/pymbar/pull/302 merged in, but if you're planning on having pymbar 4 released very soon, then maybe its not worth his effort.

zhang-ivy:
@mrshirts : I had a discussion today with @jchodera  and the perses / openmmtools developers -- since our tools depend on pymbar3, we are willing to take over maintaining pymbar3, so I think @mikemhenry will work on cutting a bugfix release that fixes the CI and contains the bootstrap branch you wrote awhile ago (https://github.com/choderalab/pymbar/pull/302). 

I'll make sure to carefully compare the code in this branch with that of the older branch and will fix any bugs in the old branch. I will also test both branches to make sure they yield similar results. Will post an update once I've done this!

mrshirts:
I have concerns about maintaining multiple branches.  I think that it's fine to keep pymbar3 alive for a while to make sure that toolchains have time to adapt, and port any bugfixes back, but I think it's better to keep new development linear.  The output formats of pymbar3 are really restrictive in addition.

Note that #302 is incorrect in at least one way (randomly selects from all samples, instead of ones from one point), and it doesn't support bootstrapping of expectations, etc.  

@zhang-ivy, shoot me an email at Colorado address and we can set up a time to talk over things with @jchodera / perses developers?  I'd rather allocate time in helping people to move forward rather than supporting old, out-of-date code.

mikemhenry:
@zhang-ivy are you going to make a new PR with just the bootstrapping stuff, or take overt this one? I just want to make sure either on this PR or a future PR, the correct branch `pymbar3` is selected.

mrshirts:
Yes, so what we want is once #302 is merged in to `master`, and a 3.0.8 version is released, then afterwards `master` should point to `pymbar4`, and what is currently master should be `pymbar3_ltm` or something like that.  @zhang-ivy , @jchodera , is that your understanding?

mrshirts:
The failing histogram test just seems to be that a histogram with random samples is not getting data in that histogram.  I will check over that problem again - maybe the way it's handled needs to be tweaked so this is less likely to happen.

zhang-ivy:
>Yes, so what we want is once https://github.com/choderalab/pymbar/pull/302 is merged in to master, and a 3.0.8 version is released, then afterwards master should point to pymbar4, and what is currently master should be pymbar3_ltm or something like that. @zhang-ivy , @jchodera , is that your understanding?

Yes!

mrshirts:
@Lnaden, thoughts about adjusting the pymbar tests relating to uncertainties?  Specifically, what would you look at for the fact that we can't just add 'bootstrap' as an uncertainty_method variable, since MBAR needs to be initialized with bootstraps for this to happen? So therefore, when `uncertainty_method = bootstrap` is called, we need to  those MBAR objects to be initialized with bootstrapping - but we don't want ALL of them to be initialized with bootstraps, because that takes a lot of extra computational time.

> Currently, to get bootstrap uncertainties from various functions, as one will now want to check both the default uncertainties and the bootstrap uncertainties, and MBAR needs to be initialized differently each way, which the pytests don't seem to handle

pymbar/fes.py
```diff
@@ -2415,16 +2416,20 @@ def ddexpf(x, k_inner):
                             h[i, j] += pE / pF[k]
 
         elif spline_weights == "unbiasedstate":
+
+            def ddexpf(x, index_i, index_j):
```
mrshirts, line 2420: I think the function is supposed to be inside of the loop, as the function (if I recall!) need to be redefined with the loop.  I remember struggling with the right way to do this. Now, there's probably a better way to do it than I did, but I will need to go back and make sure the behavior hasn't changed.   Before release at least.

mrshirts: Ah, I see what you did.  OK, I think that will work, I will double check to make sure.  It probably improves performance . . .

Lnaden:
> thoughts about adjusting the pymbar tests relating to uncertainties? 

A couple of options. The simplest is to do nothing special and just initialize the `n_bootstraps` on the `mbar_and_test` fixture. Because the `mbar_and_test` fixture is a `scope=module` fixture, it only ever gets initialized once for the whole module of tests per parameter, a total of twice. At `n_bootstrap=1000`, this adds about 15 seconds to the total test suite time on my local machine.

There is a separate option where we can rebuild the tests to pass bootstrap parameters, but I think that adds far more complexity than is needed here.

Let me do a quick tinkering to test bootstrap uncertainty method and to add bad initialization options to make sure things fail correctly as well.

mrshirts:
> it only ever gets initialized once for the whole module of tests per parameter, a total of twice.

Ah, I didn't realize that! 

And 1000 bootstraps are not needed.   200 should be plenty for decent uncertainties (so 3 seconds?) - maybe even 50, but then you might get a few statistically off samples.

mrshirts:
So, it's not doing the automated testing because there's now a merge conflict (I changed the base from the "pymbar4 at the time of fork", to "pymbar4 now", which I'm not sure was the correct thing to do, correct?

mrshirts:
Also @zhang-ivy I noticed that 'BAR' initialization is just set up for the very beginning not as one of the solvers.  I'm not sure if would help THAT much to start from BAR (versus starting from the previous step) but I can test and see (though in pymbar 4, by moving that into a separate function that can be called.

Lnaden:
> pymbar4 at the time of fork", to "pymbar4 now", which I'm not sure was the correct thing to do, correct?

What you probably wanted to do was `rebase` this, or not have changed it at all. Git is pretty good about 3-way merge (normal) and then fastforwarding commits. Since we're already here though, don't do anything else, we just have to fix the merge conflict.

mrshirts:
OK, so the following does appear to make bootstrapping faster (If the states overlap in order).  See the latest commit for the code.

1. Initialize each bootstrap iteration with the "original" f_k.
2. The run BAR (with the bootstrapped energies), starting from these original f_k.  BAR gives results that are closer (at least in some representative cases). 
3. Pass these to adaptive, without having a minimum number of self-consistent iteration.
4. Set tolerance to 10-6. 

At least for one relatively difficult data set, it seems to take only 1-3 iterations per bootstrap. 

I can potentially put the changes in #302 if you can in exchange do some more hard testing on pymbar4 :)

mrshirts:
"You" is @zhang-ivy who wants the features, not @Lnaden!

mrshirts:
> How many iterations does are required per bootstrap for that difficult dataset if you don't use any of the 4 improvements you mentioned?

For this particular data set, ~1500 steps for adaptive, or a run of L_BFGS + 2.

Without bar, just initializing from the true number, 4-5 steps.  requiring 2 steps of self-consistent iteration (which helps adaptive converge), then 4-7 steps.  Increasing the tolerance adds a step usually, but there's a decent difference between 3 and 2 steps!  

> Sounds good. Can you elaborate on what you mean by hard testing?

Try it on the variety of calculations you would use pymbar3 for, basically, plus trying a few combinations of inputs for each one.

zhang-ivy:
@mrshirts : I just realized I don't think I understand `min_sc_iter` in this branch. I thought it was telling the solver the minimum number of self consistent iterations to run. However, it doesn't seem to actually control whether the self consistent iteration is computed (as both self consistent and newton raphson are computed either way [here](https://github.com/choderalab/pymbar/blob/pymbar4_bootstrapping/pymbar/mbar_solvers.py#L350-L358)) -- it only controls which one is chosen. Do you think it would make more sense to refactor this so that if `sci_iter < min_sc_iter`, the self consistent gradient/free energy is not computed? I would guess this would speed things up quite a bit.

mrshirts:
Sorry for not being clear.  It does computes both, but requires it to select the self_consistent_iteration to be chosen for the first `min_sc_iter` iterations. This is because in rare situations, NR diverged.  A couple of SC seemed to always resolve the problem.   That is an empirical finding, though.

mrshirts:
> @mrshirts : Ok, so why would we expect that skipping self consistent iterations would speed things up? If both self consistent and NR are being each iteration computed either way, then we wouldn't expect skipping the sc iterations to speed anything up, right?

If we are far away, then SC is safer, but close, it is significantly slower.  If we are close to the answer, then each NR iteration should be a better solution, and should converge faster, and SC iterations are wasted.  This is the theory, and it appear to be true in practice - finishing in 1-3 iterations, instead of 3-5.

mrshirts:
> Even though we know that for bootstrapping, we only want the NR solution, we are wasting computational effort each time we compute the SC solution. 

We don't actually know that!

> While I haven't actually timed how long the self consistent computation takes, I assume its a bottleneck

It involves fewer calculations than NR, so is faster; I have not timed to see how much faster (1/2 the time, maybe?).

zhang-ivy:
>I don't think it's a good idea, as if the bootstrap DOES happen to be too far out, you don't necessarily want it to crash. That loses a lot more time that the simulation going slowly. SC will pull it back if it goes off. It's somewhat more important it be robust (and completes) than it be fast (and crashes 1% of the time).

Ah, makes sense. Thanks!

> Can you send me an example data file, and I can play around with what should speed it up?
Here are np arrays of the data I used to generate these [results](https://github.com/choderalab/pymbar/pull/302#issuecomment-1183600742). The only problem is that github doesn't allow me to attach files > than a certain limit so I cut the number of samples in half (shape of u_kn is now (36, 90000)) and saved the arrays for you.
[bootstrapping_input.zip](https://github.com/choderalab/pymbar/files/9106155/bootstrapping_input.zip)

>It involves fewer calculations than NR, so is faster; I have not timed to see how much faster (1/2 the time, maybe?).

Ah good to know! My guess was wrong then, thanks for clarifying.

mrshirts:
With pymbar3 (out of the box), I'm getting 188 seconds for 100 bootstraps.  I tried switching to "L-BFGS-B" for the bootstrapped solver, and it took 194 second, so pretty comparable. 

With pymbar4, I'm getting 80 seconds to do 100 bootstraps.   Here, we are skipping the forced SC iterations, and have a larger tolerance (10^-6 vs 10^-12), and I think that's the main effect.   That seems to be a little faster that you are getting, but I don't know why. 

Interestingly, for this system, the MBAR solutions for the "original" data are closer than the the BAR solutions for the bootstrapped sample, which is interesting. What is the degree over overlap for this data set?  My other dataset perhaps had less overlap, so the initialization was not as good. 

I think using a larger tolerance for bootstrap and removing the "force-take-SC" on the initial steps will help.

This is getting dangerously into the area of optimizing both pymbar3 and pymbar4 simultaneously, which gets messy.   At a certain point, it gets easier to use that time to switch to pymbar4. 

I think there are some fundamental limits to how fast it has to go since one has to set things up, etc. **What amount of time do you need it to take**, to understand what is necessary?  Note that 50 bootstraps is probably decent for computing standard deviations. 

Now, pymbar4 will have jax as well, so that may help, we will see (still need to merge those branches, so nothing above reflects it).

zhang-ivy:
>With pymbar3 (out of the box), I'm getting 188 seconds for 100 bootstraps. I tried switching to "L-BFGS-B" for the bootstrapped solver, and it took 194 second, so pretty comparable.

Whoa, that's super fast. When I run the default pymbar3 bootstrapping for 100 bootstraps, it takes 565 seconds. When I run 10 bootstraps, it takes 66 seconds. I'm running on a GTX 1080Ti with the following code:

```python

with open("u_kn.npy", "rb") as f:
    decorrelated_u_ln = np.load(f)['arr_0']
    
with open("n_k.npy", "rb") as f:
    decorrelated_N_l = np.load(f)['arr_0']

initial_time = time.time()
mbar = MBAR(decorrelated_u_ln, decorrelated_N_l, nbootstraps=100, solver_tolerance=1e-12)
results = mbar.getFreeEnergyDifferences(compute_uncertainty=True, uncertainty_method='bootstrap', return_dict=True)
print(results['Delta_f'][0, -1], results['dDelta_f'][0, -1])
print("Time: ", time.time() - initial_time)

```
Note I added `solver_tolerance` as an argument to the API but haven't pushed to #302 yet.

>Interestingly, for this system, the MBAR solutions for the "original" data are closer than the the BAR solutions for the bootstrapped sample, which is interesting. What is the degree over overlap for this data set? My other dataset perhaps had less overlap, so initialization with BAR helps. I will switch back to pymbar4 and look at this a bit more.

The overlap is likely pretty poor as the trajectory represents a relative binding free energy calculation for mutating an interfacial residue in a protein:protein system (barnase:barstar). 

>I think using a larger tolerance for bootstrap and removing the "force-take-SC" on the initial steps will help.

Yep, agreed. I'm working on making the changes to #302 now.

> This is getting dangerously into the area of optimizing both pymbar3 and pymbar4 simultaneously, which gets messy. At a certain point, it gets easier to use that time to switch to pymbar4.

Agreed! 

> What amount of time do you need it to take, to understand what is necessary? Note that 50 bootstraps is probably decent for computing standard deviations.

I don't have a good answer to this -- the longer the analysis takes, the more time it'll take for me to generate plots, and I assume I'll have to re-do this many times as I write up my manuscript, so I was just trying to see how much I could speed it up by playing around with things for a day or so.  I'm happy with where we're at and like you said, I don't think we want to spend any more effort optimizing pymbar3 anyway.

As for the number of bootstraps necessary -- this was something I had been thinking about today -- I had assumed we would need 1000 bootstraps to be confident in our uncertainty estimate, but maybe that's not necessary. Why do you think 50 is enough?

mrshirts:
For this particular data set, BAR is having a hard time converging - it's hitting the maximum iterations.  Not sure why, but that explains why it doesn't do so well.  I will investigate later, but I don't know that that's really the medium term solution for the speedup.

mrshirts:
> Whoa, that's super fast. When I run the default pymbar3 bootstrapping for 100 bootstraps, it takes 565 seconds. When I run 10 bootstraps, it takes 66 seconds. I'm running on a GTX 1080Ti with the following code:

I'm just running on my MaBook Pro laptop. Maybe it uses the cores effectively? I tried rerunning the exact code you just sent (same settings) on pymbar 3 , #302 PR, and it completed in 128 seconds for 100 bootstraps (that was on the shorter 36x90000 matrices you sent). 

> As for the number of bootstraps necessary -- this was something I had been thinking about today -- I had assumed we would need 1000 bootstraps to be confident in our uncertainty estimate, but maybe that's not necessary. Why do you think 50 is enough?

Brad Ephron's book says 50-200 is enough for a good covariance.  If you wanted a rough estimates, I think 20 should be plenty, then use 200 for final results.   The uncertainty might be off by a factor of 50% or so with just 20 samples, but it's enough to give you a basic understanding of what is going on.   1000 is almost certainly overkill.

zhang-ivy:
>I'm just running on my MaBook Pro laptop. Maybe it uses the cores effectively? I tried rerunning the exact code you just sent (same settings) on pymbar 3 , https://github.com/choderalab/pymbar/pull/302 PR, and it completed in 128 seconds for 100 bootstraps (that was on the shorter 36x90000 matrices you sent).

Oh my bad, I really don't need to be running this on a GPU, since its not GPU accelerated. I just used a CPU and requested 24 threads and it runs MUCH faster, like 3-4x faster, which is what I was originally aiming for. Thanks for helping me catch that mistake!

>Brad Ephron's book says 50-200 is enough for a good covariance. If you wanted a rough estimates, I think 20 should be plenty, then use 200 for final results. The uncertainty might be off by a factor of 50% or so with just 20 samples, but it's enough to give you a basic understanding of what is going on. 1000 is almost certainly overkill.

Ah ok good to know, thanks! I'll see if I can get my hands on his book to learn more about this but will use your suggestion moving forward.

mrshirts:
> I just used a CPU and requested 24 threads and it runs MUCH faster, like 3-4x faster, which is what I was originally aiming for.

There we go!

> Ah ok good to know, thanks! I'll see if I can get my hands on his book to learn more about this but will use your suggestion moving forward.
https://www.hms.harvard.edu/bss/neuro/bornlab/nb204/statistics/bootstrap.pdf

mrshirts:
I made the "use BAR to initialize each bootstrap" more robust, but in the end, it wasn't as fast as just initializing from the original guess. I'll think about the best logic!

mrshirts:
Ugh, getting the issue again where black locally doesn't give the same results as black in github for the four files, so I don't know what to fix.

mrshirts:
Hmm, I think the lint issue was actually with pymbar4, not pymbar4_bootstrapping.  Trying again.

mrshirts:
Going to merge this now, and figure out lint later.

mrshirts:
> Any idea why this is happening? I would've thought it'd be resolved by https://github.com/choderalab/pymbar/pull/443 but then I realized this error occurs at line 730 in mbar.py, and the PR fixes a later line.

The fix was in #441, #443 was a bugfix to #441, so maybe it's still #441?

mrshirts:
One other question that would be great to know on a hard system - how does the pymbar4 bootstrapping work for entropy and enthalply errors?  (n_bootstraps in __init__ and uncertainty_method = "bootstrap").

zhang-ivy:
>Is it returning a successful result in the end? Then it is working as intended (it tried 'hybr', then something else). Maybe there should be an additional message saying what method succeeded?

Yes, its returning a successful result. I think adding an additional message with the method that succeeded would be nice!

> The fix was in https://github.com/choderalab/pymbar/pull/441, https://github.com/choderalab/pymbar/pull/443 was a bugfix to https://github.com/choderalab/pymbar/pull/441, so maybe it's still https://github.com/choderalab/pymbar/pull/441?

Just saw this comment: https://github.com/choderalab/pymbar/pull/441/files#diff-39154a13f665187718c90bf1b7a14a9edb282028dd0fbd94a8dfb02168e42580R648-R652

Do you think we should change `logfactor` to small number like 0.001? 

>One other question that would be great to know on a hard system - how does the pymbar4 bootstrapping work for entropy and enthalply errors? (n_bootstraps in initializations and uncertainty_method = "bootstrap" in compute_entropy_and_enthalpy).

Sure I can check this later tonight!

mrshirts:
> Do you think we should change logfactor to small number like 0.001?

In the current code, it's something like 4x(the smallest mantissa accessible to numpy, like 1e-15).  That has worked so far.  0.001 is not so good, since then numbers near in magnitude 0.001 can't be averaged.

mrshirts:
> Sure I can check this later tonight!

Great!

zhang-ivy:
@mrshirts : I just tested bootstrapping for `compute_free_energy_differences()` and `compute_entropy_and_enthalpy()` with this code (with pymbar4 only, since pymbar3 doesn't do bootstrapping for `compute_entropy_and_enthalpy()`):

```python
    # Test compute_free_energy_differences()
    decorrelated_u_ln, decorrelated_N_l = analyzer._compute_mbar_decorrelated_energies()
    mbar = MBAR(decorrelated_u_ln, decorrelated_N_l)
    results = mbar.compute_free_energy_differences(compute_uncertainty=True)
    print("compute_free_energy_differences()", results['Delta_f'][0, -1], results['dDelta_f'][0, -1])
    
    # Test compute_entropy_and_enthalpy()
    results = mbar.compute_entropy_and_enthalpy()
    print("compute_entropy_and_enthalpy()", results['Delta_f'][0, -1], results['dDelta_f'][0, -1], results['Delta_u'][0, -1], results['dDelta_u'][0, -1], results['Delta_s'][0, -1], results['dDelta_s'][0, -1])
    
    # Test compute_free_energy_differences() with bootstrapping
    mbar = MBAR(decorrelated_u_ln, decorrelated_N_l, n_bootstraps=200)
    results = mbar.compute_free_energy_differences(compute_uncertainty=True, uncertainty_method='bootstrap')
    print("compute_free_energy_differences() bootstrapped", results['Delta_f'][0, -1], results['dDelta_f'][0, -1])
    
    # Test compute_entropy_and_enthalpy() with bootstrapping
    results = mbar.compute_entropy_and_enthalpy(uncertainty_method='bootstrap')
    print("compute_entropy_and_enthalpy() bootstrapped", results['Delta_f'][0, -1], results['dDelta_f'][0, -1], results['Delta_u'][0, -1], results['dDelta_u'][0, -1], results['Delta_s'][0, -1], results['dDelta_s'][0, -1])
    
```

Here are the results.

For case 1 (barnase:barstar mutation):

```
compute_free_energy_differences() -52.62531138448979 0.04448291789808753
compute_entropy_and_enthalpy() -52.62531138449739 0.044482917898087226 -58.26693039969541 9.695425671134956 -5.64161901519401 9.693498249036294
Failed to reach a solution to within tolerance with hybr: trying next method
compute_free_energy_differences() bootstrapped -52.62531138448979 0.10343092206869589
compute_entropy_and_enthalpy() bootstrapped -52.62531138449739 0.10343092206869589 -58.26693039969541 69.68171363279014 -5.64161901519401 69.67825680667228
```

For case 2 (ala dipeptide mutation):
```
compute_free_energy_differences() -52.27795853969293 0.056782925355530894
compute_entropy_and_enthalpy() -52.277958539692946 0.05678292535553127 -57.77404736274184 1.8697858250872512 -5.496088823048922 1.8701373837035669
Failed to reach a solution to within tolerance with hybr: trying next method
compute_free_energy_differences() bootstrapped -52.27795853969293 0.16964559913815944
compute_entropy_and_enthalpy() bootstrapped -52.277958539692946 0.16964559913815944 -57.77404736274184 26.242413812942953 -5.496088823048922 26.239604799661123
```

Bootstrapping for free energy differences looks good (as I already found in my tests last week), but the bootstrapped uncertainty for the entropy and enthalpy looks much larger than expected...

zhang-ivy:
>In the current code, it's something like 4x(the smallest mantissa accessible to numpy, like 1e-15). That has worked so far. 0.001 is not so good, since then numbers near in magnitude 0.001 can't be averaged.

I just tried copying this line from pymbar4 mbar.py: 
```python
logfactor = 4.0 * np.finfo(np.float64).eps
        # make sure all results are larger than this number.
        # We tried 1 before, but expecations that are all very small (like
        # fraction folded when it is low) cannot be computed accurately.
        # 0 causes warnings in the test with divide by zero, as does 1*eps (though fewer),
        # and even occasionally 2*eps, so we chooose 4*eps
```
to replace `logfactor` [here](https://github.com/choderalab/pymbar/blob/master/pymbar/mbar.py#L648-L652).

But I'm still seeing the same warning: 

```
/home/zhangi/miniconda3/envs/perses-paper/lib/python3.9/site-packages/pymbar-3.0.5-py3.9.egg/pymbar/mbar.py:777: RuntimeWarning: divide by zero encountered in log
  Log_W_nk[:, sa] = np.log(A_n[i, :]) + Log_W_nk[:, l]
```

mrshirts:
> but the bootstrapped uncertainty for the entropy and enthalpy looks much larger than expected...

Interesting, can you post the data set somewhere I can look at it?  (like google drive?)  It was looking OK for my tests, but maybe there's an error that didn't show up with that relatively easy test.

mrshirts:
> But I'm still seeing the same warning with pymbar3:

Yeah, I'd have to take a look at the data set to see.

zhang-ivy:
The data for these experiments isn't too big, so i'm just going to attach them here. I included the `u_kn` and `n_k` for barnase:barstar mutation as well as for ala dipeptide mutation.
[pymbar3_divide_by_zero.zip](https://github.com/choderalab/pymbar/files/9145584/pymbar3_divide_by_zero.zip)

zhang-ivy:
Note that barnase:barstar and ala dipeptide have 24 and 12 thermodynamic states, respectively, but I've added unsampled endstates for lambda = 0 and lambda = 1 such that the states are ordered like [unsampled endstate for lambda 0, sampled endstate for lambda 0, ... sampled endstate for lambda = 1, unsampled endstate for lambda = 1], so the matrices have K = 26 and 14, respectively. Not sure this matters, but just noted it in case it does.

mrshirts:
Hmm.  I can't seem to load them.  When I unzip and run I get:

with open("u_kn.npy", "rb") as f:
    u_kn = np.load(f)['arr_0'] 

I get: ValueError: Cannot load file containing pickled data when allow_pickle=False

And when I try 
 u_kn = np.load(f,allow_pickle=True)['arr_0']

I get:
*** _pickle.UnpicklingError: Failed to interpret file <_io.BufferedReader name='u_kn.npy'> as a pickle

Any thoughts?

zhang-ivy:
Hmm, this is working for me: 
```python
with open("bnbs/u_kn.npy", "rb") as f:
    u_kn = np.load(f, allow_pickle=True)

with open("bnbs/n_k.npy", "rb") as f:
    n_k = np.load(f, allow_pickle=True)
```

mrshirts:
Hah, figured out what was doing wrong. Back to work!

mrshirts:
Ah, I figured out what went wrong with the expectations bootstrapping. A couple of arrays that needed to be bootstrapped were not. It appears to give very close results now.

zhang-ivy:
@mrshirts : Nice! 

Did you have a chance to look into why this warning is appearing for pymbar3 and not pymbar4? 
```
/home/zhangi/miniconda3/envs/perses-paper/lib/python3.9/site-packages/pymbar-3.0.5-py3.9.egg/pymbar/mbar.py:777: RuntimeWarning: divide by zero encountered in log
  Log_W_nk[:, sa] = np.log(A_n[i, :]) + Log_W_nk[:, l]
```

If not, no worries, I will look into that this week.

mrshirts:
I'll try to take a look later today.

mrshirts:
There's one really weird point - 24090 is much lower than the other points, and is giving problems.  Not sure if that affects things.

mrshirts:
So it fails because with the logic, there always be a number that is zero.  So the idea is when subtracting the minimum value for each row, you actually subtract (minimum_value - small amount), which will result of all entries in that line to be greater than zero.  So setting a slightly larger value for logfactor should solve the problem.

zhang-ivy:
>There's one really weird point - 24090 is much lower than the other points, and is giving problems. Not sure if that affects things.

Can you clarify what you're referring to when you say `24090`? And what specifically is lower? the free energy?

zhang-ivy:
>So it fails because with the logic, there always be a number that is zero. So the idea is when subtracting the minimum value for each row, you actually subtract (minimum_value - small amount), which will result of all entries in that line to be greater than zero. So setting a slightly larger value for logfactor should solve the problem.

Hmm I set logfactor to ` logfactor = 4.0 * np.finfo(np.float64).eps` in my local version of pymbar3's `bootstrap` branch and I'm still seeing the same warning..

mrshirts:
Ah, OK, I understand better now, and why it was solved in pymbar4. When we subtract out the minimum, in order for it not to be zero, we need to subtract np.min(A) + some factor such that (A-np.min(A))/np.min(A) is greater than eps.  This logic is fixed in pymbar 4, but not pymbar 3 - i.e. the factor we subtract is not too small in absolute terms, but not in relative terms; so technically A-(A_min-eps) = eps for the lowest value, but that is not enough, since if A is too big, then A-(A_min-eps) will round to 0.   See improved logic in pymbar 4 on https://github.com/choderalab/pymbar/blob/e8a68ff0a6e9ef56cf5afed08f5090679e20fefd/pymbar/mbar.py#L872 and the next few lines.

I don't fully understand why the code was not working for very small expectations (expectations of occupation probabilities that were much lower than 1), but it clearly was not, and the new changes seem to fix this.  I will continue to test with the new code to see if there are any issues we didn't get to.

mrshirts:
Note that the particular data set is not great, because there's one data point that is 1400 units below the rest of the data points, leading to highly likely overflow.

zhang-ivy:
Thanks for the explanation! I ported the pymbar4 changes you mentioned to my local branch of pymbar3 and the divide by 0 warning goes away. 

>But at least you see why the pymbar 4 one does work, and can decide how much to port over.

It seems like its a pretty easy fix to port the changes over (only involves changing a few lines), but I'm curious how bad is it to see the divide by 0 warning? In other words, are the returned expectations wrong when we see the warning? I would assume they aren't, but i'm not sure what the implications of dividing by 0 are on the resulting expectations.

mrshirts:
> but I'm curious how bad is it to see the divide by 0 warning? In other words, are the returned expectations wrong when we see the warning? I would assume they aren't, but i'm not sure what the implications of dividing by 0 are on the resulting expectations.

My understanding is that give the error, the point that is zero (which is the minimum) is left out, which may or many not affect the results.  If the lowest point is very non-representative it would affect it a lot, if it was relatively representative then it would not affect things much.

zhang-ivy:
Ah I see. It sounds like it would definitely be good to port the changes to pymbar3 then. I'll open a new branch for that fix.

PR #449, opened 2022-07-05
Getting CI working on the current default branch

mikemhenry:


mikemhenry:
@mrshirts It looks like a precision issue/something about the test that makes it non-deterministic, since it passes for one of the jobs. Thoughts?

mrshirts:
The sigmas from BAR and MBAR should not actually be exactly equal.  More specifically, BAR has 2 equations for asymptotic uncertainty that both converge to the same value, but approach it slightly differently for smaller N. Currently, it's using the one that is NOT exactly the same as MBAR.  See the comments in the pymbar4 version of BAR (in other_estimators.py)

So the test is too constrained.  Saying the two sigmas are the same to within 1% is fine for now.

mikemhenry:
@mrshirts Okay I think I rewrote the test correctly https://github.com/choderalab/pymbar/pull/449/commits/40fc7a011a47b28c289f5cfabb27ad2554034470

mikemhenry:
@mrshirts just a friendly bump to review this PR :)

mrshirts (approved):
I am not an expert on the testing framework, but I don't see any actual code changes that would cause problems.

jchodera (approved):
@mikemhenry Thanks so much for getting this working! This looks great!

After merging this, we will probably want to rename `master` to `pymbar3` or similar so we can do long-term maintenance releases on this branch. We'll also need to make sure to rename the branch in the GHA workflows that reference `master` if GitHub doesn't do this for us automagically.

mikemhenry:
> if GitHub doesn't do this for us automagically.

It won't, but now I've updated the PR, good catch!

PR #451, opened 2022-07-08
Convert JAX JIT to Decorators

Lnaden:
This PR makes the `jax.jit` calls all decorators where used. This makes no effort to be efficient `jit` calls beyond the one conditional code path in `jax_self_consistent_update`.

This PR also makes all the tests pass under this setup, and should get the GitHub Action CI file to correctly pull the JAX (cpu) install. GPU install is beyond the scope of what GHA can reasonably be expected to do.

Fixed an issue in testing where newer version of numpy's `assert_almost_equal` failed because multi-value array default truth is ambiguous. Replaced with `assert_array_almost_equal`, but we should consider replacing with the more stable `assert_allclose`, but it has different tolerances.

Lastly, ensured the `Deltaf_ij` and `dDeltaf_ij` are normal NumPy arrays, which are mutable, instead of the JAX DeviceArray objects which are immutable. This only maters for in-place update calls.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -31,7 +31,7 @@ jobs:
 
     steps:
     - name: Checkout the code
-      uses: actions/checkout@v1
+      uses: actions/checkout@v3
 
     - name: Additional info about the build
       shell: bash
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -9,6 +9,8 @@ dependencies:
   - numpy >=1.12
   - scipy
   - numexpr
+  - jaxlib
+  - jax
 
     # Testing
   - pytest
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -598,8 +598,11 @@ class MBAR:
             None,
         )  # By default, returns None for dDelta and Theta
 
-        # Compute free energy differences.
-        Deltaf_ij = self.f_k - np.vstack(self.f_k)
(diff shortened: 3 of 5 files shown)


mrshirts:
Thanks for the patience on responses on this - I've been teaching at a workshop.   

I guess what I was originally wondering if it would be possible to have it have a non-jax version.  I tried uninstalling jax, and it didn't run (ie. the line @jax.jit triggered "name 'jax' is not defined".)  

But if jax is robustly conda installable, maybe we don't need a non-jax version, just make sure it runs without a GPU installed.  And if jax installs, we can probably assume that jit will work. 

This is the decision point I'm just not sure I understand that well.   Oliver seem to be OK with JAX (see https://github.com/alchemistry/alchemlyb/discussions/205).

Lnaden:
Therein lies part 2 of the plan from #447. I've broken this transition up into smaller, easier to review chunks. This PR simply converts things to decorators, the second PR after it will allow JAX or nor JAX with no duplicated code really. I've got that implemented on my branch here: https://github.com/lnaden/pymbar/tree/overlap_no_jax

There is some code duplication and the code paths are not perfectly exact due to the difference in L-BFGS and BFGS that are coded in, but that can be cleaned up later too.

mrshirts:
> Therein lies part 2 of the plan from https://github.com/choderalab/pymbar/pull/447. I've broken this transition up into smaller, easier to review chunks. This PR simply converts things to decorators, the second PR after it will allow JAX or nor JAX with no duplicated code really.

OK, great - this seems to work fine on tests and examples, so  I will go ahead and merge it, and will start looking at the other  branch.

PR #454, opened 2022-07-13
Align namespaces to work if JAX is not installed

Lnaden:
This sets all of the imports of JAX to align with the standard NumPy and Scipy imports so all the code continues functioning when JAX is not installed or has an issue installing/importing.

Adds a passthrough identity decorator so if `jax.jit` is unavailable, the decorator just returns the function, doing nothing.

This does create a few odd names, but they should be clear enough to function without issue.

Added a new CI variable to test mbar without JAX installed.

Because `use_jit` becomes `False`, the codepaths with the decorator may not be tested properly yet.

This is part 2 of #447 improvements I listed

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -24,6 +24,7 @@ jobs:
       matrix:
         os: [macOS-latest, ubuntu-latest, windows-latest]
         python-version: [3.8, 3.9]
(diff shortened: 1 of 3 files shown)


mrshirts:
This runs on examples, and changes make sense, so merging.

PR #457, opened 2022-07-18
Remove old_mbar.py

mrshirts:
Remove old_mbar.py - not needed anymore.


mikemhenry:
I think you missed this one: https://github.com/choderalab/pymbar/blob/pymbar4/pymbar/tests/test_covariance.py#L40

And I'm not sure if we care about these or if they will be removed as well someday:

https://github.com/choderalab/pymbar/blob/pymbar4/scripts/misc/time_expectations.py#L18
https://github.com/choderalab/pymbar/blob/pymbar4/scripts/benchmark_mbar.py#L27-L28

mrshirts:
I think the scripts should be moved to the pymbar_datasets repository (https://github.com/choderalab/pymbar-datasets).  Doing extensive benchmarking really requires having a lot of datasets that one can test on, but we don't want to carry those datasets along with the code.  Also, these scripts are broken since the datasets are now in this separate data repository anyway.  I think I'll go ahead and move them now . . .

mrshirts:
OK, how does this look?  I've removed those tests and the dataset scripts, as well as some code that was only used in removed code.

Issue #468, opened 2022-07-20
Fix RTD documentation build for 3.0.7 / pymbar3-ltm


mrshirts:

mikemhenry:
Builds now!
https://pymbar--471.org.readthedocs.build/en/471/index.html
Working on these fixes here: 
https://github.com/choderalab/pymbar/pull/471

PR #471, opened 2022-07-27
sphinx.ext.pngmath is now sphinx.ext.imgmath

mikemhenry:


mikemhenry:
Resolves #468

mikemhenry:
Okay it isn't perfect but it looks way better now, just need to update the README and add a warning

PR #485, opened 2022-12-20
Do not use removed `numpy.bool`

mattwthompson:
```python3
>>> import numpy
>>> numpy.__version__
'1.24.0'
>>> numpy.bool
<stdin>:1: FutureWarning: In the future `np.bool` will be defined as the corresponding NumPy scalar.  (This may have returned Python scalars in past versions.
  File "<stdin>", line 1, in <module>
  File "/Users/mattthompson/mambaforge/envs/openff-interchange-env/lib/python3.9/site-packages/numpy/__init__.py", line 284, in __getattr__
    raise AttributeError("module {!r} has no attribute "
AttributeError: module 'numpy' has no attribute 'bool'

### DIFF
--- a/examples/parallel-tempering-2dfes/parallel-tempering-2dfes.py
+++ b/examples/parallel-tempering-2dfes/parallel-tempering-2dfes.py
@@ -212,7 +212,7 @@ else:
 # ===================================================================================================
 
 # Create a list of indices of all configurations in kn-indexing.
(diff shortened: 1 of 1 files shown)


mattwthompson:
The docs builds are broken, which is not related and should be fixed elsewhere

I think this is good to go

mikemhenry (approved):

Issue #493, opened 2023-03-15
readthdocs not building; looks like config issue


mrshirts:
readthedocs appears to be failing on build; see https://readthedocs.org/projects/pymbar/builds/19784267/ specifically seems to be behind on numpy versions (jax and jaxlib appears to need 1.20, but only 1.19.2 is installed.  I'm not quite expert enough to fix this.

mikemhenry:
Yes! I was poking at this yesterday while wrapping up pymbar 3 & 4 compatibility for openmmtools. Switching to using mamba for building the docs env will speed it up and also fix issues with using pip to create it.

mikemhenry:
I will keep this issue open until I get the lts branch merged in #495

mikemhenry:
https://pymbar.readthedocs.io/en/3.1.1/
https://pymbar.readthedocs.io/en/4.0.1/

Both working now :)

PR #494, opened 2023-03-15
switch to using mamba for rtd building

mikemhenry:
Solves #493


mikemhenry:
I will make the same fix to the LTS branch so we can get docs building on that as well

mikemhenry:
This should be good to go!

mrshirts (approved):

ijpulidos (approved):
This looks good to me. The failing tests on windows are unrelated to the changes in this PR, correct? LGTM!

PR #507, opened 2023-06-16
Correct docstring default value

badisa:
Noticed that the default value of the tolerance was off and fixed up a few typos


Lnaden (approved):
LGTM thanks! The docs are not building right now due to upstream issues that are beyond our direct control, but these are correct regardless.

PR #510, opened 2023-06-26
fixes RTD build

mikemhenry:
see https://github.com/readthedocs/sphinx_rtd_theme/issues/1463


Lnaden (approved):
Looks good. I thought the problem was on the RTD backend site which was failing builds, not a local install which is why I hadn't fixed it yet.

Other failures  here has nothing to do with what was changed here and I've got it fixed in another PR.

PR #513, opened 2023-08-23
Fixed example bug and updated output.

mrshirts:
Fixed a bug in the harmonic_oscillators.py example that wasn't actually retrieving the data from the results object for reverse exponential averaging.

Added logging to the two harmonic oscillator files.

Update the output of the example scripts since they were quite old.


mrshirts:
Anyone have any thoughts why the doc build failed and the windows test timed out?  This change would not really affect that much - it's adding logging instructions to 2 files, making one trivial change in processing outputs (doesn't change the algorithms run), and adds some text files.

mikemhenry:
@mrshirts I can look at RTD, I think the windows test is more just being flaky, I will re-run that one

mikemhenry (approved):

PR #517, opened 2023-11-07
Assorted maintenance updates

mattwthompson:
Here are a handful of things I noticed while working on #516 but aren't directly related to adding new features. I hope the commit messages are explanatory enough, but I'm happy to provide more detail as needed.


Lnaden:
I wound up implementing some of this as part of #526, but I want to keep this open to pull some other ideas when I can cycle back to this. Thanks for proposing these ideas!

PR #524, opened 2024-02-29
Support newer jax version (0.4.25) where jax.config.config is removed.

nielskm:


Lnaden:
This is implemented as part of #526. Thank you for calling this to our attention.

PR #526, opened 2024-03-21
Update to Python 3.12

Lnaden:
Closes #516 by retrying it. Easier to make a new PR for it. We'll need to do this to make PyMBAR 3.12 viable on Conda-Forge due to the versioneer changes. Python 3.8 and 3.12 pass locally.

Thanks to @mattwthompson who made the initial PR which this just uses that code directly.


Lnaden:
This also closes #524 by fixing the Jax config import. That PR would have fixed the problem for most other codes, but JAX does something weird under the hood where the `try...except` around `from jax.config import config` would wind up mangling the namespace for `config` which removed the ability to check for `x64_enabled`. The change on this PR fixes that by providing a single call which works in pre and post 0.3.25.

Lnaden:
I do also want to credit and thank @nielskm for opening the PR which called attention to the JAX issue.

PR #535, opened 2024-08-26
Fix to bootstrap of free energy surfaces, affecting timing

mrshirts:
Free energy surface code was calling MBAR after each call to randomizing bootstraps.  It does not appear to affect the results, but slows things down by a factor of a little less than 2x (146 vs 87 seconds for one sample run).

### DIFF
--- a/pymbar/fes.py
+++ b/pymbar/fes.py
@@ -399,11 +399,11 @@ class FES:
                         0, N_k[k], size=N_k[k]
                     )
                     index += N_k[k]
-                    # recompute MBAR.
-                    mbar = pymbar.MBAR(
-                        self.u_kn[:, bootstrap_indices], self.N_k, initial_f_k=self.mbar.f_k
-                    )
-                    x_nb = x_n[bootstrap_indices]
+                # recompute MBAR.
+                mbar = pymbar.MBAR(
+                    self.u_kn[:, bootstrap_indices], self.N_k, initial_f_k=self.mbar.f_k
+                )
+                x_nb = x_n[bootstrap_indices]
 
             # Compute unnormalized log weights for the given reduced potential
             # u_n, needed for all methods.


mrshirts:
Couple more changes I will put in to fix uncertainties, don't do anything yet to it . . .

mrshirts:
OK, I think I got the changes in I needed to.

pymbar/fes.py
```diff
@@ -399,11 +399,11 @@ def generate_fes(
                         0, N_k[k], size=N_k[k]
                     )
                     index += N_k[k]
-                    # recompute MBAR.
-                    mbar = pymbar.MBAR(
-                        self.u_kn[:, bootstrap_indices], self.N_k, initial_f_k=self.mbar.f_k
-                    )
-                    x_nb = x_n[bootstrap_indices]
+                # recompute MBAR.
```
mrshirts, line 402: This was unnecessary - it was running MBAR too many times.  This saves approximately 2X time.

pymbar/fes.py
```diff
@@ -1415,8 +1417,8 @@ def _get_fes_histogram(
                 fall = np.zeros([len(histogram_data["f"]), n_bootstraps])
                 for b in range(n_bootstraps):
                     h = histogram_datas[b]  # just to make this shorter
-                    fall[:, b] = h["f"] - h["f"][j]
-                df_i = np.std(fall, axis=1)
+                    fall[:, b] = h["f"] - h["f"][j]  # subtract out the reference bin
+                df_i = np.std(fall, ddof=1, axis=1)
```
mrshirts, line 1421: Fixing the std definition.

pymbar/fes.py
```diff
@@ -1510,7 +1512,7 @@ def _get_fes_histogram(
                         fall[i, j, b] = (
                             histogram_datas[b]["f"] - histogram_datas[b]["f"].transpose()
                         )
-                dfxij_vals = np.std(fall, axis=2)
+                dfxij_vals = np.std(fall, ddof=1, axis=2)
```
mrshirts, line 1515: Fixing std definition

pymbar/fes.py
```diff
@@ -692,8 +692,9 @@ def _generate_fes_kde(self, b, x_n, w_n):
             params = self.kde.get_params()  # pylint: disable=access-member-before-definition
             kde.set_params(**params)
         else:
-            kde = self.kde
-        kde.fit(x_n, sample_weight=self.w_n)
+            kde = self.kde  # use these new weights for the KDE
+            w_n = self.w_n
```
mrshirts, line 696: I actually can't remember if this was 100% necessary to get updated weights . . .

pymbar/fes.py
mrshirts, line 1421: Fix bootstrap std definition

pymbar/fes.py
```diff
@@ -1565,9 +1567,15 @@ def _get_fes_kde(
         if reference_point == "from-lowest":
             fmin = np.min(f_i)
             f_i = f_i - fmin
+            wheremin = np.argmin(
```
mrshirts, line 1570: Need to find the location that this is zeroed at for the actual computation of the std.

pymbar/fes.py
```diff
@@ -1565,9 +1567,15 @@ def _get_fes_kde(
...
+                f_i
+            )  # needed or uncertainties, as we zero by location, not values
         elif reference_point == "from-specified":
             fmin = -self.kde.score_samples(np.array(fes_reference).reshape(1, -1))
             f_i = f_i - fmin
+            wheremin = np.argmin(
```
mrshirts, line 1576: Need to find the location that this is zeroed at for the actual computation of the std.

pymbar/fes.py
```diff
@@ -1594,8 +1602,9 @@ def _get_fes_kde(
 
             fall = np.zeros([len(x), n_bootstraps])
             for b in range(n_bootstraps):
-                fall[:, b] = -self.kdes[b].score_samples(x) - fmin
-            df_i = np.std(fall, axis=1)
+                fall[:, b] = -self.kdes[b].score_samples(x)
+                fall[:, b] -= fall[wheremin, b]
```
mrshirts, line 1606: Zero out at the correct location.

mrshirts (commented):
My comments on this for other people.

mrshirts:
Suggestions for anyone else who should review - or if there's anyone who could take a look?  We are looking at some free energy surface problems for OpenFF, so we want to get this through.

PR #539, opened 2024-08-27
Use self.u_kn when verbose=True in MBAR

### LINKED ISSUE #538: Verbose=True with MBAR results in an error for alternative input types of u_kn
A small improvement in consistency within pymbar would improve usability in alchemlyb.

### Backgound
Although the input, u_kn, for MBAR is defined to be a np.ndarray, it is converted to an array in line 239 with:
`        self.u_kn = np.array(u_kn, dtype=np.float64)`
Packages (i.e., alchemlyb) then instead will pass a pd.DataFrame and find that all works as expected due to this conversion.

### Problem
When `MBAR.fit(verbose=True)` is used, there is an indexing error if the provided u_kn is a pd.DataFrame. This is because of line 296:
`                    uzero = u_kn[k, indices] - u_kn[l, indices]`
where u_kn is used instead of self.u_kn.

### Solution
If `self.u_kn` were used in this line instead of `u_kn`, the users of alchemlyb would be able to benefit from the option, `verbose=True`.

jaclark5:
Resolves #538

Line 296 of mbar.py is changed from:
`                    uzero = u_kn[k, indices] - u_kn[l, indices]`
to
`                    uzero = self.u_kn[k, indices] - self.u_kn[l, indices]`

### DIFF
--- a/pymbar/mbar.py
+++ b/pymbar/mbar.py
@@ -293,7 +293,7 @@ class MBAR:
             for k in range(K):
                 for l in range(k):
                     diffsum = 0
-                    uzero = u_kn[k, indices] - u_kn[l, indices]
+                    uzero = self.u_kn[k, indices] - self.u_kn[l, indices]
                     diffsum += np.dot(uzero, uzero)
                     if diffsum < relative_tolerance:
                         self.samestates.append([k, l])


mikemhenry:
@jaclark5 Can you add a test that does the behavior you mention in #538 That will help us prevent any regressions in the future (and might catch some edge case where the numpy array vrs pandas data frame issue crops up)

mrshirts:
There are some weird cases where one needs either self.u_kn or a copy of it, but this should be totally fine because its just doing some analysis of whether the states are the same or not, not manipulating anything.

jaclark5:
> @jaclark5 Can you add a test that does the behavior you mention in #538 That will help us prevent any regressions in the future (and might catch some edge case where the numpy array vrs pandas data frame issue crops up)

Sure, can you help me picture what that would look like? Since `MBAR.fit()` sets `self.u_kn = np.array(u_kn, float)` I don't foresee a similar issue unless one changes the `MBAR.fit()` method, as u_kn will be lost and only the attribute, `MBAR.u_nk` will be pertinent to other methods.

mikemhenry:
@jaclark5 you are right!

mikemhenry (approved):

mikemhenry:
The linting errors are happening on master, so this one is good to merge in!

PR #549, opened 2025-03-07
Clarify `n_bootstraps` docstring

IAlibay:
The docstring seems to indicate `Optional[int]` but the code expects `int`. This clarifies the docstring a little bit by removing mentions of `None`.


mrshirts:
Yep, you're right, it was changed so that the default is zero, not none.

mikemhenry:
Since this doesn't use pre-commit, the linter is ran on the whole code base instead of just the changed files (or the diff) -- but it passes black so this LGTM

mikemhenry (approved):

PR #550, opened 2025-03-15
Correct Defaults in MBAR Docstring and Fix Typos

badisa:
@proteneer recently observed that the MBAR class's docstring had some inconsistencies in the default values. This PR resolves those inconsistencies as well as fixes up any random typo I found in the codebase.


jchodera (approved):
Wow, this is fantastic! Thanks so much for the contribution!

In a separate issue, we should address whether the default `MBAR` initializer parameter `n_bootstraps` should be `0` (current) or `None`.

jchodera:
The failing [lint-format](https://github.com/choderalab/pymbar/actions/runs/13876653504/job/38829814621?pr=550) check appears to be due to the fact that the `mamba-org/provision-with-micromamba@main` action is no longer maintained:
```
Run mamba-org/provision-with-micromamba@main
...
Warning: This action is deprecated and no longer maintained. Please use mamba-org/setup-micromamba instead. See `https://github.com/mamba-org/provision-with-micromamba#migration-to-setup-micromamba` for a migration guide.
```
with the real error being a `-p` argument provided to `micromamba-root`
```
  /home/runner/micromamba-bin/micromamba shell init -s zsh -p /home/runner/micromamba-root -y --log-level warning
  The following argument was not expected: -p
```
@mikemhenry : Can you help fix that CI issue in a separate PR?

mrshirts:
Typo fixes look good.  Lotta typos there over a lot of files over time . . .

PR #551, opened 2025-03-19
Make MBAR bootstrapping deterministic independent of verbose flag

badisa:
Noticed while reading through the code recently that the determinism of the MBAR class bootstrapping changes depending on whether or not `verbose` is enabled. This was surprising, as I expected to be able to disable verbose and get the same results. This PR resolves that inconsistency and verifies determinism.

Also switches to using a [Numpy Random generator](https://numpy.org/doc/stable/reference/random/generator.html) to use the latest numpy convention as well as avoid impacting the global RNG (though still queries the global to get a seed if none is provided).

Feel free to close this or let me know how I can make it better. It does now generate a random sequence on construction that may or may not be used, but seemed less surprising than having determinism change based on a verbose flag.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macOS-latest, ubuntu-latest, windows-latest]
-        python-version: ["3.8", "3.11", "3.12"]  # Check against oldest and newest versions
+        python-version: ["3.10", "3.11", "3.12", "3.13"]  # Check against oldest and newest versions
         jax: ["", "_jax"]
         exclude:  # Skip win + jax
           - jax: "_jax"
@@ -100,7 +100,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: ["3.10"]
     env:
       PACKAGE: "pymbar"
 
--- a/devtools/conda-envs/test_env.yaml
+++ b/devtools/conda-envs/test_env.yaml
@@ -1,12 +1,11 @@
 name: test
 channels:
   - conda-forge
-  - defaults
 dependencies:
     # Base depends
   - python
   - pip
-  - numpy >=1.12
+  - numpy >=1.12,<2.3
   - scipy
   - numexpr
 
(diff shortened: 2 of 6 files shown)


mrshirts:
Thanks for the catch that it was using random numbers within the verbose, which would change the behavior. 

I'm hesitate to move ALL of that outside the verbose flag verbose, since I recall we had some test cases where this took quite some time.  Could something be done like storing the state of the random number generator? 

Changes to improve the random number generator are welcome, but maybe that should be a separate commit?

mikemhenry:
Thanks for this! I will make sure there are no performance regressions in the tests @mrshirts 

It looks like the windows platform is failing, I think the fix is to make the dtype explicit -- I can make that change

mrshirts:
Hmm, yes, the actual loops are in the verbose.  It seems sort or weird coding-wise to generate a bunch of random numbers and them throw them away when verbose is not selected, but I guess it doesn't have any real performance hits.

mikemhenry:
I am not sure how else to do it since when you call  `indices = self.rng.choice(np.arange(self.N), maxpoint)` you will affect the state of the RNG, so if it is moved inside the `verbose` switch, then even with the same seed you will produce different random numbers since you are calling `self.rng` a different number of times.

mrshirts:
> I am not sure how else to do it since when you call `indices = self.rng.choice(np.arange(self.N), maxpoint)` you will affect the state of the RNG, so if it is moved inside the `verbose` switch, then even with the same seed you will produce different random numbers since you are calling `self.rng` a different number of times.

IF we don't see a way to do it otherwise, I'm OK with this.

mikemhenry (approved):

mikemhenry:
@ijpulidos do you have any ideas on a better way to do this?

ijpulidos:
Apologies for the late response, I think this approach works. I really appreciate using the "new"  `default_rng()` which makes it better to handle and pass around if needed. I think this is exactly a case where this new API shines, so I'd say we are using it correctly. Maybe we do need to now make sure that we are depending on `numpy>=1.17` since this is the version where the new API was introduced, iirc. Other than that, I this looks good to me. Thanks for the contribution!

PR #558, opened 2025-07-02
Fix issue 552

### LINKED ISSUE #552: IndexError in subsample_correlated_data in non-conservative mode
A potential `IndexError` has been identified in the `subsample_correlated_data` function, specifically within the non-conservative subsampling logic (when `conservative` is `False`).

The issue arises in the `while` loop where indices for the uncorrelated subsample are determined. The loop counter `n` increments in each iteration, but new indices are only appended to the `indices` list if the calculated time index `t` is different from the last added index. If consecutive calculated `t` values are duplicates, `n` can become larger than the number of elements in `indices`. Subsequently, the condition `(n == 0) or (t != indices[n - 1])` can attempt to access `indices[n - 1]` with an out-of-bounds index, leading to an `IndexError`.

**Steps to Reproduce (Example Scenario):**

While a precise minimal example might require further investigation into the behavior of `statistical_inefficiency`, a scenario that could potentially trigger this is when the statistical inefficiency `g` is very small, or when the `int(round(n * g))` calculation results in repeated values of `t` for multiple consecutive values of `n`.

Here's a conceptual example that might illustrate the issue:

```python
import numpy as np
# Assuming statistical_inefficiency would return a small value
g = np.float32(1.0013024806976318)
A_t = np.random.rand(19000) # Example time series
indices = []
n = 0
T = len(A_t)
while int(round(n * g)) < T:
    t = int(round(n * g))
    if (n == 0) or (t != indices[n - 1]):
        indices.append(t)
    else:
        print('Warning: duplicate index')
    n += 1
```

In this simplified loop (mimicking the relevant part of the function), there is an index the error when `n - 1` exceeds the bounds of `indices`.

**Expected Behavior:**

The function should return a list of indices representing an uncorrelated subsample of the data without raising an `IndexError`.

**Actual Behavior:**

Under certain conditions, an `IndexError` can be raised due to an attempt to access `indices[n - 1]` where `n - 1` is not a valid index.

**Suggested Fix:**

Modify the `if` condition within the non-conservative subsampling block to ensure that `indices[n - 1]` is only accessed when `indices` is non-empty and its length is at least `n`. A safer approach is to compare with the last element using `indices[-1]` after checking if the list has any elements.

Replace the line with:

```python
if (n == 0) or (t != indices[- 1]):
```

mikemhenry:
Prevented indexerror

Fixes #552

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -100,7 +100,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.10]
+        python-version: ["3.10"]
     env:
       PACKAGE: "pymbar"
 
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macOS-latest, ubuntu-latest, windows-latest]
-        python-version: ["3.8", "3.11", "3.12"]  # Check against oldest and newest versions
+        python-version: ["3.10", "3.11", "3.12", "3.13"]  # Check against oldest and newest versions
         jax: ["", "_jax"]
         exclude:  # Skip win + jax
           - jax: "_jax"
@@ -100,7 +100,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [3.8]
+        python-version: [3.10]
     env:
       PACKAGE: "pymbar"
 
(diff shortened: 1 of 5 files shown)


mikemhenry:
@mrshirts Does this seem like a good fix for #552

mrshirts:
I haven't checked in in detail, but it seems like this should work.  If this solves the problem, I don't think it introduces extra ones.

PR #566, opened 2025-10-06
Test newer versions of numpy

mikemhenry:


mikemhenry:
If we get another all green pass here then we can relax the numpy version on the conda-forge feedstock

PR #569, opened 2025-10-10
disable jax if PYMBAR_DISABLE_JAX is set

mikemhenry:
This will let a user disable jax at runtime. A larger refactor is needed to test this since at import time the `PYMBAR_DISABLE_JAX` variable is checked, so even if monkey patched the module has already been imported. This also means that once pymbar has been imported, setting `PYMBAR_DISABLE_JAX` will do nothing.

~TODO: add docs~

### DIFF
--- a/docs/moving_from_pymbar3.rst
+++ b/docs/moving_from_pymbar3.rst
@@ -168,7 +168,7 @@ install the appropriate GPU libraries). ``jax`` will be installed when
 or without ``jax`` installed if there are issues with the JAX configuration.
 
 To disable the ``jax`` acceleration, set the environment variable 
-``PYMBAR_DISABLE_JAX`` to ``TRUE``. This will disable ``jax`` acceleration
+``PYMBAR_DISABLE_JAX=TRUE``. This will disable ``jax`` acceleration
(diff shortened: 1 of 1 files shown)


pymbar/mbar_solvers.py
```diff
@@ -11,7 +12,16 @@
...
-force_no_jax = False  # Temporary until we can make a proper setting to enable/disable by choice
+
+
+# Temporary until we can make a proper setting to enable/disable by choice at runtime
+def _setup_jax_accleration():
+    return os.environ.get("PYMBAR_DISABLE_JAX", "").lower() in ("true", "yes", "0")
+
+
+# Setup if we should use jax or not
+force_no_jax = _setup_jax_accleration()
```
ijpulidos:
```suggestion
force_no_jax = _setup_jax_acceleration()
```

Fix typo

ijpulidos (changes requested):
This is looking great, I think this works for our purposes. Just some minor changes/comments.

In the long term I wonder if we want to make the `MBAR` object itself specify the backend via an argument.

PR #571, opened 2026-02-14
Use MBAR Relative Tolerance in solvers

badisa:
While running some simulations I noticed that despite passing `relative_tolerance=1e-7` to the MBAR object, the  warnings about convergence indicated that the tolerance was set to `1e-12`. This PR makes sure to pass the relative tolerance all the way down to the `solve_mbar` method.


mrshirts:
Thanks for seeing this.  We are updating a few more things for the solver, so we may wait a bit before integrating this to make sure we integrate everything together.

Issue #574, opened 2026-04-25
Chunked / streaming u_kn for memory-efficient MBAR fits at large K


jamesoliverh:
# Chunked / streaming `u_kn` for memory-efficient MBAR fits

## Summary

`pymbar.MBAR` requires the full `(K, N)` reduced-potential matrix `u_kn` to be materialized in RAM before fitting. For large multi-state problems this becomes the binding memory constraint long before any other resource is exhausted, blocking otherwise-tractable fits on commodity workstations. Adding a streaming / chunked `u_kn` provider would give a roughly **20–40× memory reduction** at modest (5–30%) per-iteration cost — and on memory-pressured systems would actually *speed up* total fit time by avoiding swap-thrashing.

## Concrete numbers (motivating example)

I'm running a multi-histogram FS / MBAR fit for a thermodynamic-geometry problem (2D Ising, L=96) with:
- K = 309 thermodynamic states (Phase-1 anchor MC at distinct (β, h) points)
- 30,000 thinned independent samples per state → N = 9.27M total samples
- `u_kn` = 309 × 9.27M × 8 B = **23 GB** for the matrix alone.
- Internal weight buffer `W_nk` (also `(N, K)`) doubles peak memory to ~46 GB if materialized.

This exceeds the available RAM on the host (16 GB) and forces either:
1. Decimation/thinning at the sample level (information loss; some structural details of the surface get washed out)
2. Swapping (10× slowdown from disk I/O per iteration)
3. Dropping pymbar in favour of a hand-rolled chunked solver (loses pymbar's well-tested adaptive-Newton convergence machinery and BAR-chain initialization)

None of these are great. The natural fix is to chunk `u_kn` along the sample axis.

## Why this is fixable

The MBAR fixed-point recursion and gradient/Hessian evaluations are **sums over samples**. Each per-iteration computation can be re-expressed as

```
G = Σ_t g(u_kn[:, t], state)
```

where `g(...)` returns a per-sample contribution to the gradient/Hessian/log-denominator. So the natural API is:

- An `iter_u_kn_chunks(batch_size)` callable that yields `(K, B)` slices.
- The solver loops over chunks, accumulating into shared `(K,)` and `(K, K)` buffers.
- Each chunk fits in `K × B × 8` bytes; with `B = 100k`, peak chunk memory at K=309 is ~250 MB regardless of total N.

Memory drops from `O(K × N)` to `O(K × B + K²)` — for our case from 23 GB to ~250 MB.

[...]

PR #575, opened 2026-04-26
Chunked + JAX-jit'd MBAR with vectorised batch evaluation

jamesoliverh:
This PR adds chunked working-array MBAR end-to-end across the fit and the post-fit batch evaluation. This caps peak working memory at `O(chunk_size × K)` regardless of sample count `N`, so MBAR can be fit and evaluated on datasets that previously exceeded available RAM. A single `chunk_size` knob streams the fit's inner reductions over the sample axis, and on the evaluation side replaces the per-target Python loop with one vectorised pass per chunk, skipping the dense `(N, K + NL + S)` log-weight allocation. JAX-jit'd per-chunk inner kernels run throughout (numpy fallback when JAX isn't available), measuring around 2× faster per chunk than the numpy path. Default behaviour without `chunk_size` is unchanged. A streaming-callable form for `u_kn` is also supported, so out-of-core fits don't need to load the full matrix into RAM.

Closes issue #574.

### LOC summary

| Section | Impl LOC | Notes |
|---|---|---|
| 1. End-to-end chunking (`_chunked.py`) | ~290 | Whole new module (322 LOC) |
| 2. Vectorised batch evaluation (`mbar.py`) | ~40 | `chunk_size` kwarg + chunked branch + 3 wrapper kwargs |
| 3. JAX work (`_chunked.py` kernels + `mbar_solvers.py` bypass) | ~70 | 4 `@jit` kernels + numpy fallback + JIT-bypass guard |
| **Implementation total** | **~400** | |
| Tests | ~625 | 5 test files |
| **Cumulative PR diff** | **+1149 / -19** | |

### Three areas of work

1. **End-to-end chunking.** New `pymbar/_chunked.py` module holds the chunking primitives and per-chunk kernels. The fit's inner reductions and the batch-evaluation reductions both route through it under one `chunk_size` knob.
2. **Vectorised batch evaluation.** `compute_expectations_inner` and the three public wrappers gain a `chunk_size` kwarg. When set, all target states are evaluated in one vectorised pass per chunk; the per-target Python loop and the dense `(N, K + NL + S)` allocation are gone on this branch. Default `chunk_size=None` runs the existing path bit-for-bit unchanged. `return_theta=True` and `uncertainty_method='bootstrap'` are rejected on the chunked path — neither composes cleanly with column-chunked kernels.
3. **JAX work.** Four `@jit`-decorated per-chunk kernels in `_chunked.py`, with a clean numpy fallback (`@jit` reduces to identity, `jnp` becomes `np`) when JAX is unavailable or `PYMBAR_DISABLE_JAX=1`. The outer `jax.jit` that wraps the adaptive solver is bypassed when chunking is active, since the chunked helpers have side-effecting Python loops that can't be traced.

### Tests

Unit tests for the new chunked helpers against `scipy.special.logsumexp`. Integration tests for the three public methods against the dense reference at `decimal=10`, parametrised over a range of `chunk_size`. Constraint tests for the `return_theta=True` / bootstrap rejection paths. A regression test that forces the adaptive solver and clears the JAX cache to exercise the JIT-bypass cold trace. 131 tests pass, 9 xfailed pre-existing, no regressions.

### Memory / performance

For workloads where `N ≫ chunk_size` and `K + NL + S` is large, target-axis work moves from a Python loop to one vectorised reduction per chunk, and peak memory drops by `O(N / chunk_size)`. JAX-jit'd per-chunk kernels measure around 2× faster than the numpy path.

### Out of scope (follow-ons if requested)

- Streaming `u_ln` via a callable yielding column chunks (motivation weaker — `u_ln` is usually constructed cheaply from a target-state list).
- `Theta` from chunked weights (would need a redesign of `_computeAsymptoticCovarianceMatrix`).
- Bootstrap with chunked path (feasible by re-running chunked helpers per replicate, not in this PR).

### Notes for review

The PR is broader than issue #574's original scope: the chunked path covers fit + post-fit eval under one `chunk_size` knob, with JAX-jit'd per-chunk kernels throughout. If splitting the batch-eval helpers and the `compute_*` changes off into a separate PR would help review, I can rebase that piece onto its own branch.

### DIFF
--- a/pymbar/_chunked.py
+++ b/pymbar/_chunked.py
@@ -1,0 +1,137 @@
+"""Chunked working-array variants of the MBAR inner-fit operations.
+
+The default `mbar_solvers.py` calls
+
+    log_denominator_n = logsumexp(f_k - u_kn.T, b=N_k, axis=1)
+
+which materializes a (N, K) working array the same size as ``u_kn``. For
+large N, this is the dominant peak-memory cost during the fit -- not
+``u_kn`` itself. See pymbar issue #574.
+
+These helpers compute the same quantities by streaming over the sample
+axis, with peak working-array memory ``O(chunk_size * K)`` instead of
+``O(N * K)``. Numpy-only; the JAX path is preserved for callers that fit
+in memory.
+"""
+import numpy as np
+
+
+def chunked_log_denominator(u_kn, N_k, f_k, chunk_size):
+    """Streaming logsumexp(f_k - u_kn.T, b=N_k, axis=1).
+
+    log_denom[n] = log sum_k N_k * exp(f_k - u_kn[k, n])
+    """
+    N = u_kn.shape[1]
+    with np.errstate(divide='ignore'):
+        log_N = np.log(N_k).astype(u_kn.dtype)  # -inf for N_k = 0; drops cleanly
+    a = (f_k + log_N).astype(u_kn.dtype)  # (K,)
+    out = np.empty(N, dtype=u_kn.dtype)
+    for s in range(0, N, chunk_size):
+        e = min(s + chunk_size, N)
+        z = a[:, None] - u_kn[:, s:e]  # (K, B)
+        m = z.max(axis=0)
+        np.subtract(z, m, out=z)
+        np.exp(z, out=z)
+        out[s:e] = m + np.log(z.sum(axis=0))
+    return out
+
+
+def chunked_log_numerator_k(u_kn, log_denominator_n, chunk_size):
+    """Streaming logsumexp(-log_denominator_n - u_kn, axis=1).
+
+    log_num_k[k] = log sum_n exp(-log_denominator_n[n] - u_kn[k, n])
+
+    Maintains a running max-shift per state across chunks.
+    """
+    K, N = u_kn.shape
+    m_k = np.full(K, -np.inf, dtype=u_kn.dtype)
+    S_k = np.zeros(K, dtype=u_kn.dtype)
+    for s in range(0, N, chunk_size):
+        e = min(s + chunk_size, N)
+        z = -log_denominator_n[s:e][None, :] - u_kn[:, s:e]  # (K, B)
+        m_chunk = z.max(axis=1)
+        m_new = np.maximum(m_k, m_chunk)
+        # exp(m_k - m_new) is 0 when m_k = -inf (initial); guard the all-inf
+        # corner case where m_new is also -inf.
+        with np.errstate(invalid='ignore'):
+            scale = np.where(np.isfinite(m_new), np.exp(m_k - m_new), 0.0)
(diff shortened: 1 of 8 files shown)


mrshirts:
Being able to deal with larger matrices could be important. A few thoughts:

* I am worried about making MBAR less maintainable.  There are limited (near zero?) resources for maintenance and we are behind as it is on robustness issues.  This would complicate the code in a way that would likely make it less maintainable
* The PR reads as though it was AI-generated (Formatting, sentence structure, level of detail).  I'd like @jamesoliverh to comment a bit more on this and walk through the logic? I wasn't able to figure out who this is with 5 min on his repositories.
* The "right" way to do this is to not have to construct the entire MBAR u_kn matrix, but stitch together overlapping locally solved states, but that work has been stuck for 15 years . . .

jamesoliverh:
Hi, thank you for your comments / concerns.

Indeed the PR was generated with the help of Opus 4.7. I trust that's not an issue? 

I'm actively using this package for a project and used Opus to draft an improvement that gets me around a significant bottleneck and imho this represents a significant improvement of the functionality and ability to scale to larger problems, centred around chunking a) the internal operations, for significantly less memory use and (surprisingly, maybe) increased speed, and b) the u_nk matrix itself. 

I'm also working on the JAX implementation and vectorised batch evaluation post-fit with chunking (again a significant improvement in my workflow), to be added shortly.

I'm open to all feedback and will iterate on it. If you want to know more about me, my Google Scholar is here:

https://scholar.google.com/citations?user=vYV9IgQAAAAJ&hl=en

and LinkedIn is here: 

https://www.linkedin.com/in/james-hamp-327260126?utm_source=share&utm_campaign=share_via&utm_content=profile&utm_medium=ios_app 

(not sure what else you might be looking for in terms of info about me).

I have contributed (pre Gen AI!) to numpy and various other libs. 

Thanks very much for commenting and really look forward to working together to get this functionality into the lib.

mrshirts:
Thanks for the information on the background!

> Indeed the PR was generated with the help of Opus 4.7. I trust that's not an issue?

I don't there's an issue per se, but there is the worry that if the submitter may not understand the logic, and there could be some underlying changes that we don't realize that are happening, since AI can occasionally hallucinate, as I have certainly see in my own reviewing of code by Copilot.

We do need an official AI code policy!

mikemhenry:
Nothing intrinsically wrong using AI (well maybe some ethical issues with training and resource consumption) but it is easy to produce a huge output `+1,149` that can be challenging for a human to read. This project already has limited resources for maintenance. A lot of the code added is tests and it looks like ~620 lines of non-test code.

I am happy to take a look at it!  

I haven't looked at it yet so this might exist in some doc strings, but can you add a little example on how to use it/enable it?

